PR #3, opened 2018-10-09
Remove what lives in TopologicalInventory engine

agrare:
Models and migrations live in the topological_inventory engine (https://github.com/agrare/topological_inventory)


agrare:
@bdunne was committing `db/schema.rb` a design choice?

bdunne:
> was committing db/schema.rb a design choice?

No, I don't think so

agrare:
> No, I don't think so

Okay cool I'll go ahead and remove it

PR #4, opened 2018-10-09
Ignore database.yml since it can vary for devs

bdunne:
In production we're using ENV["DATABASE_URL"]


carbonin:
Are you going to make the same change on the persister side?

carbonin:
@agrare You good with these? Don't want to confuse local dev too much

agrare:
:+1: works for me

PR #7, opened 2018-10-09
Remove ProvidersController and move SourcesController to v0.0

bdunne:


app/controllers/api/v1x0.rb
```diff
@@ -1,5 +1,4 @@
 module Api
   module V1x0
-    class SourcesController < Api::V1::SourcesController; end
   end
 end
```
agrare: Do we need this at all if there is nothing in V1x0?

bdunne: Not yet.  Removed.

PR #12, opened 2018-10-10
Add container_projects API

gtanzillo:


public/doc/swagger-2-v1.0.0.yaml
```diff
@@ -17,6 +17,47 @@ security:
   - UserSecurity: []
 
 paths:
+  /container_projects{container_project_id):
```
bdunne: Should be `/container_projects/{container_project_id}`

gtanzillo: Good eye, thanks!

gtanzillo:
Closing because the original fork from agrare was deleted

PR #18, opened 2018-10-15
Update swagger doc and add a routes spec

bdunne:


public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -3,134 +3,714 @@ info:
...
+      - application/json
+      responses:
+        200:
+          description: Endpoints array
+          schema:
+            type: array
+            items:
+              "$ref": "#/definitions/Endpoint"
+    post:
+      summary: Create a Endpoint
```
gtanzillo: This would read better if it were "Create an EndPoint". Same for the ones that follow.

carbonin: Why? The class is called `Endpoint` and "endpoint" is a single word, not two.

carbonin: Oh, "a" vs "an" should be fixed though. Didn't catch that one.

gtanzillo: Yeah, sorry for the confusion. I was referring to the "a" vs "an"

gtanzillo: Nice! I was able to fix the typo in my comment 🤓

bdunne: Fixed

PR #21, opened 2018-10-18
Add Subcollections

bdunne:
- Add subcollections for existing things that can be subcollections
- Remove unnecessary test helper methods by changing `:type => :controller` to `:type => :request`
- Refactor almost identical code for Controller.list, .show, .destroy into mixins
- Test the mixins themselves rather than the controllers to ensure that the same cases are covered for all controllers.


agrare:
Looks great :+1: 

```
http://localhost:3000/api/v0.0/service_offerings/367/service_parameters_sets
[{"id":"362","source_id":1,"source_ref":"43d3e23d214c26dbebc0879e44425db4","name":"default","description":"An APB that deploys MediaWiki","service_offering_id":367,"extra":null,"created_at":"2018-10-17T17:15:47.070Z","updated_at":"2018-10-17T20:59:29.441Z","source_deleted_at":null,"tenant_id":1,"source_created_at":"2018-08-31T00:35:53.000Z","create_json_schema":{"type":"object","$schema":"http://json-schema.org/draft-04/schema","required":["mediawiki_db_schema","mediawiki_site_name","mediawiki_site_lang","mediawiki_admin_user","mediawiki_admin_pass"],"properties":{"mediawiki_db_schema":{"type":"string","title":"Mediawiki DB Schema","default":"mediawiki","pattern":"^[a-zA-Z_][a-zA-Z0-9_]*$"},"mediawiki_site_lang":{"type":"string","title":"Mediawiki Site Language","default":"en","pattern":"^[a-z]{2,3}$"},"mediawiki_site_name":{"type":"string","title":"Mediawiki Site Name","default":"MediaWiki","pattern":"^[a-zA-Z]+$"},"mediawiki_admin_pass":{"type":"string","title":"Mediawiki Admin User Password"},"mediawiki_admin_user":{"type":"string","title":"Mediawiki Admin User (Cannot be the same value as Admin User Password)","default":"admin"}},"additionalProperties":false},"update_json_schema":null,"archived_on":null}]
```

app/controllers/api/mixins/destroy_mixin.rb
```diff
@@ -0,0 +1,12 @@
+module Api
+  module Mixins
+    module DestroyMixin
+      def destroy
+        model.destroy(params[:id])
```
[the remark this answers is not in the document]
Fryguy: You can probably do it right here too...something like

`params.require(:id)` instead of `params[:id]`

spec/controllers/api/v0x0/endpoints_controller_spec.rb
```diff
@@ -43,12 +18,13 @@
   end
 
   it "post /endpoints creates an Endpoint" do
-    post_path(api_v0x0_endpoints_url, :body => {:host => "example.com", :source_id => source.id.to_s, :tenant_id => tenant.id.to_s}.to_json, :format => :json)
+    headers = { "CONTENT_TYPE" => "application/json" }
+    post(api_v0x0_endpoints_url, :params => {:host => "example.com", :source_id => source.id.to_s, :tenant_id => tenant.id.to_s}.to_json)
```
Fryguy: Why was this changed to :params from :body?  I would expect post params in a body.

Fryguy:
One thing that concerns me here is the idea that we were going to generate the swagger yaml from the controllers.  I had assumed we'd have to leverage code comments to fill in the documentation aspects or use something like swagger-docs to do the heavy lifting.  If the code is moved into mixins like this though, I'm not sure how that would work.

spec/controllers/api/v0x0/endpoints_controller_spec.rb
bdunne: ¯\\\_(ツ)_/¯ That's just the way request specs work in rails 5 https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec#providing-json-data

bdunne:
I have a script that updates the swagger file based on the routes, not the controllers.

PR #22, opened 2018-10-18
Use strong params

bdunne:
Fixes https://github.com/ManageIQ/topological_inventory-api/pull/21#discussion_r226353408

### DIFF
--- a/app/controllers/api/v0/sources_controller.rb
+++ b/app/controllers/api/v0/sources_controller.rb
@@ -6,20 +6,26 @@ module Api
       include Api::Mixins::ShowMixin
 
       def create
-        body = JSON.parse(request.body.read)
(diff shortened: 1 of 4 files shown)


app/controllers/api/v0/sources_controller.rb
```diff
@@ -6,21 +6,27 @@ class SourcesController < ApplicationController
...
+        Source.update(params.require(:id), update_params)
         head :no_content
       rescue ActiveRecord::RecordNotFound
         head :not_found
       end
 
       private
 
+      def create_params
+        ActionController::Parameters.new(JSON.parse(request.body.read)).permit(:name, :tenant_id)
```
Fryguy: Not for this PR, but this pattern feels like something we could provide in ApplicationController...

```ruby
def body_params
  @body_params ||= ActionController::Parameters.new(JSON.parse(request.body.read))
end
```

PR #23, opened 2018-10-18
Rename ServiceParametersSet to ServicePlan

bdunne:
Depends on https://github.com/ManageIQ/topological_inventory-core/pull/17

### DIFF
--- a/app/controllers/api/v0/service_instances_controller.rb
+++ b/app/controllers/api/v0/service_instances_controller.rb
@@ -7,7 +7,7 @@ module Api
       private
 
       def list_params
-        params.permit(:source_id, :tenant_id, :service_offering_id, :service_parameters_set_id)
+        params.permit(:source_id, :tenant_id, :service_offering_id, :service_plan_id)
       end
 
       def model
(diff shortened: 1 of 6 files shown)


agrare:
@bdunne we're not worrying about breaking API changes => bumping the major version yet right

bdunne:
No, I think we're sticking to v0.0 for now.  I briefly talked to @carbonin & @gtanzillo and think we should start versioning soon, but we have to figure out when.

agrare:
@bdunne looks like the subcollection for service_plan on service_instance isn't working:
```
No route matches [GET] "/api/v0.0/service_instances/1/service_plan"
```

bdunne:
@agrare It should be `/api/v0.0/service_instances/1/service_plans`

agrare:
@bdunne Oh okay, it is a singular belongs_to so should it be plural?

bdunne:
I see, a `ServiceInstance` `belongs_to` a `ServicePlan`, so it's not a (sub)collection, it's a sub-resource and I currently don't have any routes for sub-resources.

agrare:
Okay its not related to this PR, this looks good to me

PR #24, opened 2018-10-20
Add tests that all AR Models are serialized to match their JSONSchema

bdunne:


spec/swagger_spec.rb
```diff
@@ -15,4 +15,30 @@
   it "routes match" do
     expect(rails_routes).to match_array(swagger_routes)
   end
+
+  describe "Model serialization" do
+    let(:doc) { Api::Docs[version] }
+    let(:container_group) { ContainerGroup.create!(doc.example_attributes("ContainerGroup").symbolize_keys.merge(:tenant => tenant, :source => source, :container_node => container_node, :container_project => container_project, :source_created_at => Time.now, :source_ref => SecureRandom.uuid)) }
```
bdunne: @agrare What do you think of having some seed data and testing against that?

bdunne:
I'm going to work on the other direction (API create / update) in another PR.

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -938,6 +944,7 @@ definitions:
         format: uuid
       default:
         type: boolean
+        example: true
```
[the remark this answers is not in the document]
agrare: Hm I thought I tried setting default_value_for a while ago and it didn't actually honor it (at least not with the ruby client/sinatra server).  I'd be good with setting `default: false` in the DB for things like this.

config/initializers/application_record_serializing.rb
```diff
@@ -0,0 +1,17 @@
...
+    attrs   = previous.slice(*schema["properties"].keys)
+    schema["properties"].keys.each do |name|
+      attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
+      attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
+    end
+    attrs
+  end
+end
+
+ApplicationRecord.prepend(ar_serializing)
```
Fryguy: Why not just include this in ApplicationRecord directly?

bdunne: ApplicationRecord should live with the models in the core repo and this is very API specific.  I think that this will need to be updated to handle serializing differently for multiple versions (another thing that is not the concern of the models), so I'd rather keep it here.

PR #25, opened 2018-10-22
Prepare Travis, CodeClimate, and deployment

Fryguy:
Prepare Travis, CodeClimate, and deployment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+---
+language: ruby
+sudo: required
+cache: bundler
+rvm:
+- 2.4.5
+services:
+- docker
+addons:
+  postgresql: '9.5'
+env:
+  global:
+  - DATABASE_URL="postgresql://postgres:@localhost:5432/topological_inventory_test?encoding=utf8&pool=5&wait_timeout=5"
+before_install:
+- gem install bundler
+before_script:
+- curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+- chmod +x ./cc-test-reporter
+- "./cc-test-reporter before-build"
+- bundle exec rake db:create db:migrate
+after_script:
+- "./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT"
+deploy:
+  provider: script
+  script: bash .travis/deploy.sh
+  'on':
+    branch: master
--- a/.travis/deploy.sh
+++ b/.travis/deploy.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+TAG="$DEPLOY_SERVER/$DEPLOY_NAMESPACE/$DEPLOY_IMAGE:latest"
+
+echo $DEPLOY_TOKEN | docker login -u $DEPLOY_USER --password-stdin $DEPLOY_SERVER
+docker build -t $TAG .
+docker push $TAG
--- a/.gitignore
+++ b/.gitignore
@@ -5,12 +5,12 @@
 /.yardoc
 /Gemfile.lock
 /_yardoc/
+/bundler.d/*
 /config/database.yml
 /coverage/
+/db/schema.rb
 /doc/
 /log/
 /pkg/
 /spec/reports/
 /tmp/
-db/schema.rb
-/bundler.d/*
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,10 @@ gem 'inventory_refresh', :git => 'https://github.com/ManageIQ/inventory_refresh'
 gem 'manageiq-gems-pending', :git => 'https://github.com/ManageIQ/manageiq-gems-pending', :branch => 'master'
 gem 'topological_inventory-core', :git => 'https://github.com/ManageIQ/topological_inventory-core', :branch => 'master'
 
+group :development, :test do
+  gem 'simplecov'
+end
+
 group :test do
   gem 'rspec-rails', '~>3.8'
 end
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Swagger for Rails 5
 
+[![Build Status](https://travis-ci.org/ManageIQ/topological_inventory-api.svg)](https://travis-ci.org/ManageIQ/topological_inventory-api)
+[![Maintainability](https://api.codeclimate.com/v1/badges/47776e67dbb7cc572c3b/maintainability)](https://codeclimate.com/github/ManageIQ/topological_inventory-api/maintainability)
+[![Test Coverage](https://api.codeclimate.com/v1/badges/47776e67dbb7cc572c3b/test_coverage)](https://codeclimate.com/github/ManageIQ/topological_inventory-api/test_coverage)
+
 This is a project to provide Swagger support inside the [Ruby on Rails](http://rubyonrails.org/) framework.
 
 ## Prerequisites
@@ -23,3 +27,7 @@ To list all your routes, use:
 ```
 bin/rake routes
 ```
+
+## License
+
+This project is available as open source under the terms of the [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0).
--- a/config/database.dev.yml
+++ b/config/database.dev.yml
@@ -1,57 +1,22 @@
-# MySQL. Versions 5.0 and up are supported.
-#
-# Install the MySQL driver
-#   gem install mysql2
-#
-# Ensure the MySQL gem is defined in your Gemfile
-#   gem 'mysql2'
-#
-# And be sure to use new-style password hashing:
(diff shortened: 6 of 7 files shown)


.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+rvm:
+- 2.4.5
+services:
+- docker
+addons:
+  postgresql: '9.5'
+env:
+  global:
+  - DATABASE_USERNAME=root
+  - DATABASE_PASSWORD=smartvm
```
Fryguy: I did it this way to be consistent with service_portal-api, but @bdunne is doing is differently here: https://github.com/ManageIQ/topological_inventory-core/blob/master/.travis.yml#L13  Let's pick a way.  cc @syncrou

carbonin: I think we went that way in the other repo because we're not using `database.yml`, but here you are also avoiding duplicating the user and password in the `psql` command ... 

If we're using `database.yml` here then this looks better. And if we apply this to the core side then we can DRY the user and password.

.travis.yml
```diff
@@ -0,0 +1,29 @@
+---
+language: ruby
+sudo: required
+cache: bundler
+rvm:
+- 2.4.5
+services:
+- docker
+addons:
+  postgresql: '9.5'
```
carbonin: Might want to consider moving to a newer version of PG at some point and also moving away from the MIQ customized image. I don't think we have any need for either this version or  the changes we make to the image...

carbonin: Not for this PR though I think.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+env:
+  global:
+  - DATABASE_USERNAME=root
+  - DATABASE_PASSWORD=smartvm
+before_install:
+- gem install bundler -v 1.16.4
```
carbonin: Do we always want this version of bundler? I'll tell you right now that I'm going to forget to update this.

.travis/deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+TAG="$DEPLOY_SERVER/$DEPLOY_PROJECT/$DEPLOY_IMAGE:latest"
```
carbonin: Nit but I think the idea of the org as the project is an OpenShift concept. Is there a different name for "the second thing between slashes in the image tag" that we could use?

.travis.yml
Fryguy: Right, having a committed database.yml that's driven by the username and password env vars is what the service_portal-api does.

.travis.yml
Fryguy: I removed that and then put it back...I agree with you...let's just drop the version and install latest.

.travis/deploy.sh
Fryguy: namespace?

Fryguy:
@carbonin Updated... @bdunne Thoughts?

.travis.yml
Fryguy: What I don't like about DATABASE_URL is it overrides *all* values in database.yml including the test database.  It's not DRY with respect to database.yml in that case.  As such, local test is then different from travis test, albeit slightly

Fryguy: > too much craziness in database.yml

Which craziness?  I don't know what you mean by that.

Fryguy:
Not going to fix the Hakiri stuff in this PR.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+- docker
+addons:
+  postgresql: '9.5'
+env:
+  global:
+  - DATABASE_USERNAME=root
+  - DATABASE_PASSWORD=smartvm
+before_install:
+- gem install bundler
+- psql -c "CREATE USER $DATABASE_USERNAME SUPERUSER PASSWORD '$DATABASE_PASSWORD';" -U postgres
```
syncrou: Looking at the other repos, I like this  option best for services that commit a `database.yml`, Especially since we're DRYer with the substituted environment variables.  If we're not committing a `database.yml` the `DATABASE_URL` makes sense because that is the single declaration of `username`, `password` as well as all the other connection params.

.travis.yml
bdunne: I guess I don't know what an alternative to `DATABASE_URL` looks like that allows for a non-dirty git tree.

.travis.yml
carbonin: > If we're not committing a database.yml

I don't think you should ever be committing a `database.yml` ...

.travis.yml
Fryguy: See https://github.com/ManageIQ/service_portal-api/blob/master/config/database.yml#L5-L6

Fryguy: Note those lines can be streamlined more with

```yaml
  username: <%= ENV.fetch('SERVICE_CATALOG_DATABASE_USERNAME', 'root') %>
  password: <%= ENV.fetch('SERVICE_CATALOG_DATABASE_PASSWORD', 'smartvm') %>
```

.travis.yml
Fryguy: So, that's the general idea behind what service_portal did.  If that's *not* the direction we want to go I'm fine with that; we just need to be consistent.

.travis.yml
Fryguy: Discussed with @syncrou @bdunne and @carbonin in person, and we are going to go the route of

- not commiting database.yml
- committing a database.dev.yml with basic defaults
- changing travis.yml to just copy database.dev.yml and/or try to use travis' built-in user/database so we can avoid the psql line

Fryguy:
@carbonin Updated.

.travis/deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+TAG="$DEPLOY_SERVER/$DEPLOY_NAMESPACE/$DEPLOY_IMAGE:latest"
+
+echo $DEPLOY_TOKEN | docker login -u $DEPLOY_USER --password-stdin $DEPLOY_SERVER
+docker build -t $TAG .
```
bdunne: Should we run the tests in this container?

carbonin: Is that a thing? Wouldn't we need PG on the container or a remote PG service even if it was?

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+cache: bundler
+rvm:
+- 2.4.5
+services:
+- docker
+addons:
+  postgresql: '9.5'
+env:
+  global:
+  - DATABASE_URL="postgresql://postgres:@localhost:5432/topological_inventory_test?encoding=utf8&pool=5&wait_timeout=5"
```
carbonin: Do we even need to use the `topological_inventory_test` database here? We could drop the `db:create` and then this would be totally generic.

.travis/deploy.sh
Fryguy: no? I don't understand what you mean

The way Travis deploy section works is it
- only runs on the master branch (so on merge to master)
- only runs if the tests have passed

.travis.yml
Fryguy: There is no built-in database in Travis...only the postgres user with a blank password.  Their docs even say to do a create database yourself.

carbonin: :worried: We seem to be using a database called `travis` without creating it [in manageiq-postgres_ha_admin](https://github.com/ManageIQ/manageiq-postgres_ha_admin/blob/2ce0140a256296fb9e208fab054abdd488c456cb/spec/server_store_spec.rb#L24). [Latest build](https://travis-ci.org/ManageIQ/manageiq-postgres_ha_admin/jobs/445354598#L603) shows those specs ran, but it's entirely possible we're taking advantage of some non-public db in that case ... I can't remember if I ever saw it documented. I think @yrudman added that originally....

carbonin: @yrudman how did you find out that you could use the `travis` database?

.travis/deploy.sh
bdunne: https://docs.travis-ci.com/user/build-stages#specifying-stage-order-and-conditions

I read about someone that was doing exactly this a while back.  They had multiple stages, so they would:
- Build the container image
- Run the tests in a container from the image that was just built
- Push the image if tests passed.

It would give us the advantage that we're running tests on the same platform that we're deploying.  We could also spin up our own postgres container for test that is the same as what we run in production (we've seen differences between Travis PG and production in the past).

bdunne: Also, maybe we could tie in a QA stage and run their test suite against the same container.

.travis.yml
yrudman: I do not remember :(

Fryguy: Oh weird...I can't find info on that.  Honestly, I'd just prefer to use our own database.

PR #31, opened 2018-10-26
Add a /tasks API

agrare:
Add an API for Tasks and update swagger.yaml

```
http://localhost:3000/api/v0.0/tasks
[{"id":"1","name":"Order Service Plan","status":null,"state":null,"message":null,"progress":null,"tenant_id":1,"created_at":"2018-10-26T20:08:34.643Z","updated_at":"2018-10-26T20:08:34.643Z"}]

http://localhost:3000/api/v0.0/tasks/1
{"id":"1","name":"Order Service Plan","status":null,"state":null,"message":null,"progress":null,"tenant_id":1,"created_at":"2018-10-26T20:08:34.643Z","updated_at":"2018-10-26T20:08:34.643Z"}
```

~~Depends: https://github.com/ManageIQ/topological_inventory-core/pull/36~~

### DIFF
--- a/app/controllers/api/v0/tasks_controller.rb
+++ b/app/controllers/api/v0/tasks_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  module V0
+    class TasksController < ApplicationController
+      include Api::Mixins::IndexMixin
+      include Api::Mixins::ShowMixin
+
+      def list_params
+        params.permit(:tenant_id)
+      end
+
+      def model
+        Task
+      end
+    end
+  end
+end
--- a/public/doc/swagger-2-v0.0.1.yaml
+++ b/public/doc/swagger-2-v0.0.1.yaml
@@ -781,6 +781,41 @@ paths:
           description: Source deleted
         404:
           description: Not found
+  "/tasks":
+    get:
+      summary: List Tasks
+      operationId: listTasks
+      description: Returns an array of Task objects
+      produces:
+      - application/json
+      responses:
+        200:
+          description: Tasks array
+          schema:
+            type: array
+            items:
+              "$ref": "#/definitions/Task"
+  "/tasks/{id}":
+    get:
+      summary: Show an existing Task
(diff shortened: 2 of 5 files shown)


app/controllers/api/v0/tasks_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V0
+    class TasksController < ApplicationController
+      include Api::Mixins::IndexMixin
+      include Api::Mixins::ShowMixin
+
+      def list_params
+        params.permit(:tenant_id)
```
Fryguy: permit or require?

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -781,6 +781,41 @@ paths:
...
+      description: Returns a Task object
+      produces:
+      - application/json
+      parameters:
+      - name: id
+        in: path
+        description: ID of the resource to return
+        required: true
+        type: string
+        format: uuid
```
Fryguy: id is a uuid?

app/controllers/api/v0/tasks_controller.rb
bdunne: I think this needs to be figured out generically.  It's supposed to be in a header, right?

public/doc/swagger-2-v0.0.1.yaml
bdunne: not actually

agrare: Yeah I thought uuid was odd but every other `id` was `format: uuid` so I figured it was some swagger thing.  E.g.: https://github.com/ManageIQ/topological_inventory-api/blob/f29276eb7a40ed85642ef668cc71a17cc2236a4f/public/doc/swagger-2-v0.0.1.yaml#L38-L44

agrare:
@bdunne PTAL

PR #32, opened 2018-10-26
Customizable routes path prefix

bdunne:
Allow for ENV["PATH_PREFIX"] to change the base path for the api routes.

### DIFF
--- a/spec/swagger_spec.rb
+++ b/spec/swagger_spec.rb
@@ -11,8 +11,36 @@ describe "Swagger stuff" do
 
   let(:swagger_routes) { Api::Docs.routes }
 
-  it "routes match" do
-    expect(rails_routes).to match_array(swagger_routes)
+  describe "Routing" do
+    include Rails.application.routes.url_helpers
+
+    it "routes match" do
+      expect(rails_routes).to match_array(swagger_routes)
+    end
+
+    context "customizable route prefixes" do
+      after { Rails.application.reload_routes! }
+
+      it "with a random prefix" do
+        stub_const("ENV", ENV.to_h.merge("PATH_PREFIX" => random_path))
+        Rails.application.reload_routes!
+
+        expect(ENV["PATH_PREFIX"]).not_to be_nil
+        expect(api_v0x0_sources_url(:only_path => true)).to eq("/#{ENV["PATH_PREFIX"]}/v0.0/sources")
+      end
+    end
+
+    def words
+      @words ||= File.readlines("/usr/share/dict/words", :chomp => true)
+    end
+
+    def random_path_part
+      rand(1..5).times.collect { words.sample }.join("_")
+    end
+
+    def random_path
+      rand(1..10).times.collect { random_path_part }.join("/")
+    end
   end
 
   describe "Model serialization" do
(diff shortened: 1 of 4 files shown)


public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -14,7 +14,7 @@ securityDefinitions:
 security:
 - UserSecurity: []
 paths:
-  "/container_groups":
+  "/v0.0/container_groups":
```
Fryguy: This is  not right...this should be part of the basePath.

bdunne: Reverted

spec/swagger_spec.rb
```diff
@@ -11,8 +11,42 @@
...
-  it "routes match" do
-    expect(rails_routes).to match_array(swagger_routes)
+  describe "Routing" do
+    include Rails.application.routes.url_helpers
+
+    it "routes match" do
+      expect(rails_routes).to match_array(swagger_routes)
+    end
+
+    it "allows customizable routes" do
```
carbonin: Can you also test for the default value?

bdunne: All of the other tests test that.  If you want I can add a single expectation like we have in this test for the default case

carbonin: No this is fine then. I was confused because `api` was being removed from two tests in this PR.

bdunne: Ah.  Yeah, the test above compares the rails routes to the swagger routes and the swagger routes all start with "/api/v0.0" since that's the base path in the swagger file.

spec/swagger_spec.rb
```diff
@@ -11,8 +11,42 @@
...
+      end
+      Rails.application.reload_routes!
+    end
+
+    def random_url_char
+      (("a".."z").to_a << "_").sample
+    end
+
+    def random_path_part
+      (1..20).to_a.sample.times.collect { random_url_char }.join
```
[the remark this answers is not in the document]
Fryguy: Also simpler rand is just `rand(max)`, so in this case `rand(20).times.collect ...`

spec/swagger_spec.rb
```diff
@@ -11,8 +11,42 @@
...
+    end
+
+    def random_path
+      (1..10).to_a.sample.times.collect { random_path_part }.join("/")
```
Fryguy: Another fun way, just because...

```ruby
def random_path
  words = File.readlines("/usr/share/dict/words")
  rand(1..10).times.collect { rand(1..3).times.collect { words.sample.downcase.chomp }.join("_") }.join("/")
end
```

bdunne: good stuff...
```ruby
irb(main):012:0> puts random_path
bedrite_slabwood_heterothallism/slimsy_Rozalin_year-book_unemulative
=> nil
irb(main):013:0> puts random_path
heterogone_Strophomenacea_All-russia_Parcel/quasi-skillfully/Gallophobe_nondemocratic_Jacaltec_endocrinotherapy/perfective
```

PR #29, opened 2018-10-25
Clean up the gemfile

bdunne:
- Add version dependencies to satisfy Hakiri
- Drop unnecessary / unused dependencies

PR #33, opened 2018-10-26
This depends on the 'listen' gem

bdunne:
Followup to #29


config/environments/development.rb
```diff
@@ -43,5 +43,5 @@
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
-  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
Fryguy: I prefer we add the listen gem back in, because this is meant for do auto-reload and such.

PR #36, opened 2018-10-30
Drop dependency to gems pending

Ladas:


agrare:
If we're going to put the v2_key in `certs/` (which is a good idea) we need to set the key_root like https://github.com/ManageIQ/manageiq/blob/master/config/application.rb#L133-L134

PR #37, opened 2018-11-01
Add order endpoint for service_plans

eclarizio:
Similar to https://github.com/ManageIQ/topological_inventory-core/pull/39, I don't fully know the structure of the request that will actually be submitted to the `order` endpoint, so I've got some placeholders like `:other_params?`, and I don't know if our response is going to be taking the form of `{:task_id => task_id}` but that is what it is for now.

The specs will fail because for some reason it can't find the "order" action. I think we may need to re-think how we want to layout the routes. If another method comes up on the `ServicePlansController` that isn't a CRUD operation, we would need to add another route anyway, so maybe it's better to change the route to `service_plans/:id/order`. I'm not sure.

@agrare @gtanzillo @bdunne Please review and advise!

### DIFF
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -7,7 +7,7 @@ module Api
       def order
         service_plan = model.find(params[:service_plan_id])
         catalog_parameters = params[:service_parameters].permit!.merge(params[:provider_control_parameters].permit!)
-        task_id = service_plan.order(params[:catalog_id], catalog_parameters)
+        task_id = service_plan.order(catalog_parameters)
         render :json => {:task_id => task_id}
       end
 
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -5,8 +5,9 @@ module Api
       include Api::Mixins::ShowMixin
 
       def order
-        service_plan = model.find(params[:id])
-        task_id = service_plan.order(params[:catalog_id], params[:other_params?])
+        service_plan = model.find(params[:service_plan_id])
+        catalog_parameters = params[:service_parameters].permit!.merge(params[:provider_control_parameters].permit!)
+        task_id = service_plan.order(params[:catalog_id], catalog_parameters)
         render :json => {:task_id => task_id}
       end
 
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -5,13 +5,29 @@ module Api
       include Api::Mixins::ShowMixin
 
       def order
-        service_plan = model.find(params[:service_plan_id])
-        catalog_parameters = params[:service_parameters].permit!.merge(params[:provider_control_parameters].permit!)
+        service_plan = model.find(service_plan_id)
         task_id = service_plan.order(catalog_parameters)
         render :json => {:task_id => task_id}
+      rescue ActiveRecord::RecordNotFound
+        head :bad_request
       end
 
       private
+
+      def catalog_parameters
+        order_params[:service_parameters].merge(order_params[:provider_control_parameters])
+      end
+
+      def service_plan_id
+        params[:service_plan_id].to_i
+      end
+
+      def order_params
+        params.permit(
+          :service_parameters          => [:DB_NAME, :namespace],
+          :provider_control_parameters => [:namespace, :OpenShift_param1]
+        ).to_h
+      end
 
       def list_params
         params.permit(:source_id, :tenant_id, :service_offering_id)
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -4,6 +4,12 @@ module Api
       include Api::Mixins::IndexMixin
       include Api::Mixins::ShowMixin
 
+      def order
+        service_plan = model.find(params[:id])
+        task_id = service_plan.order(params[:catalog_id], params[:other_params?])
+        render :json => {:task_id => task_id}
+      end
+
       private
 
       def list_params
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -26,6 +26,7 @@ Rails.application.routes.draw do
       resources :source_types, :only => [:create, :index, :show] do
         resources :sources, :only => [:index]
       end
+      post "/service_plans/:id", :to => "service_plans#order"
       resources :sources,                 :only => [:create, :destroy, :index, :show, :update] do
         resources :container_groups,        :only => [:index]
         resources :container_nodes,         :only => [:index]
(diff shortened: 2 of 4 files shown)


config/routes.rb
```diff
@@ -19,6 +19,7 @@
       resources :service_plans, :only => [:index, :show] do
         resources :service_instances, :only => [:index]
       end
+      post "/service_plans/:id", :to => "service_plans#order"
```
bdunne: What does this create?  Is it a ServiceInstance?  If so, should we post to service_plans/:id/service_instances instead?

bdunne: @mkanoor @gmcculloug ^

spec/controllers/api/v0x0/service_plans_controller_spec.rb
```diff
@@ -1,4 +1,20 @@
 RSpec.describe Api::V0x0::ServicePlansController, :type => :request do
   it("Uses IndexMixin") { expect(described_class.instance_method(:index).owner).to eq(Api::Mixins::IndexMixin) }
   it("Uses ShowMixin")  { expect(described_class.instance_method(:show).owner).to eq(Api::Mixins::ShowMixin) }
+
+  describe "#order" do
+    let(:service_plan) { instance_double("ServicePlan", :id => 123) }
+
+    before do
+      allow(ServicePlan).to receive(:find).with(123).and_return(service_plan)
```
bdunne: I don't think we should stub this

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -4,6 +4,12 @@ class ServicePlansController < ApplicationController
       include Api::Mixins::IndexMixin
       include Api::Mixins::ShowMixin
 
+      def order
+        service_plan = model.find(params[:id])
+        task_id = service_plan.order(params[:catalog_id], params[:other_params?])
```
mkanoor: @eclarizio When a plan is ordered there are 2 sets of JSON parameter blobs

1. That is provider specific e.g. {'namespace' => 'my_project', 'OpenShift_param1' => 'xxxx'}
2. Plan Specific e.g. {'DB_NAME' => 'TEST_DB', 'namespace' => 'MARIA_DB_NAMESPACE1'}

We are collecting these 2 separately and sending it to the topology service

https://github.com/ManageIQ/service_portal-api/blob/db46ebe37150ac9c4f896fade228e0e2c8d91641/public/doc/swagger-2.yaml#L523

https://github.com/ManageIQ/service_portal-api/blob/db46ebe37150ac9c4f896fade228e0e2c8d91641/public/doc/swagger-2.yaml#L526

agrare: @mkanoor is the provider specific json param a part of the service plan json spec? Should it have been??

spec/controllers/api/v0x0/service_plans_controller_spec.rb
eclarizio: I tried to create the `ServicePlan` at first, but it also needs a `Source`, a `Tenant`, and a `ServivceOffering` (which also needs a `Source` and `Tenant`, though I'm guessing they would be the same ones) so I thought I'd rather have the test be fast than create a bunch of objects that (at least for this small spec) don't matter.

I do agree about not stubbing an interface we don't own, but at the same time I think it's ok to not have to worry about the database interaction for this spec. If we had a factory for it I probably would've used that but I think that's probably something reserved for another PR in the future.

bdunne: I'd much rather create the records.

Also, I'm not worried about the amount of time it takes to create the records since the whole test suite currently runs in less than 1 second.

agrare: But when we're testing with trillions of service plans... yeah lets just create the record :)

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -4,6 +4,13 @@ class ServicePlansController < ApplicationController
       include Api::Mixins::IndexMixin
       include Api::Mixins::ShowMixin
 
+      def order
+        service_plan = model.find(params[:service_plan_id])
+        catalog_parameters = params[:service_parameters].permit!.merge(params[:provider_control_parameters].permit!)
+        task_id = service_plan.order(params[:catalog_id], catalog_parameters)
```
agrare: > `params[:catalog_id]`

@mkanoor we have a `ServicePlan` belong_to a `ServiceOffering` so I think you passing in the catalog_id isn't needed

config/routes.rb
agrare: So this is still a question of if we should post to `service_plan/:id/#order` or `service_plan/:id/service_instances#create` right?  I don't feel strongly either way, I assume the latter is "more correct" from an API standpoint?

spec/controllers/api/v0x0/service_plans_controller_spec.rb
eclarizio: I know you're exaggerating with the trillions of service plans and yeah it's kind of silly cause we'll never get there, but just for fun after I changed it I did some quick timings. This spec basically goes from about ~0.3ish seconds to ~0.5ish seconds with all of the extra creates. Sure, not too big of a deal in isolation, but even just 100 tests (which I don't think is out of the realm of possibility?) and you're at an extra 20 seconds. Now, we probably don't really care *too* much because most of the time the full suite is just running on travis in the background anyway, but still, it would be really nice to be able to run the full suite locally before committing to a larger codebase and not have to wait a long time.

In the grand scheme of things, 20 seconds or even 200 seconds with 1000 tests isn't really a *huge* deal, I just don't see what actually creating the records is buying us?

I've also had to change the `allow`s a bit to now use an `allow_any_instance_of` instead of stubbing directly on the `service_plan` object because once the `.find` happens in the controller, the spec no longer has a handle to the same object that was created, and it ends up actually trying to call the `#order` method instead of simply returning the canned value.

bdunne: I'd much rather stub the `ServiceCatalogClient` than `ServicePlan.order`.  Then we're testing more than just the route.
```ruby
before do
  allow_any_instance_of(ServicePlan).to receive(:order).with(catalog_parameters).and_return("task_id")
  Rails.application.config.action_dispatch.show_exceptions = false
end
```

config/routes.rb
bdunne: cc @Fryguy Any thoughts on this?

app/controllers/api/v0/service_plans_controller.rb
mkanoor: @agrare Yes that is old stuff and not needed anymore

bdunne:
Was this ready for merge?  I don't see any changes to the swagger.yaml and still have the outstanding question about whether or not this is the correct path.

agrare:
> I don't see any changes to the swagger.yaml

Ugh you're right I assumed the specs checked for that @eclarizio can you follow up with that?

> still have the outstanding question about whether or not this is the correct path.

I assume you mean on the service_parameter/id/#order vs post to service_instance ?  I didn't see any clarification on that question for a long time and we need to get something for SSP to actually start using, figured we could come back and update it if the other way is the correct way to go.

PR #47, opened 2018-11-14
Don't .to_s ids in API specs

agrare:
https://github.com/ManageIQ/topological_inventory-core/pull/53 removed
the `h["id"] = h["id"].to_s` from the ApplicationRecord#as_json method
which is causing the specs to fail.

### DIFF
--- a/spec/controllers/api/v0x0/endpoints_controller_spec.rb
+++ b/spec/controllers/api/v0x0/endpoints_controller_spec.rb
@@ -10,12 +10,12 @@ RSpec.describe Api::V0x0::EndpointsController, :type => :request do
 
   it "post /endpoints creates an Endpoint" do
     headers = { "CONTENT_TYPE" => "application/json" }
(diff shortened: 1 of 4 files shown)


spec/controllers/api/v0x0/endpoints_controller_spec.rb
```diff
@@ -10,12 +10,12 @@
 
   it "post /endpoints creates an Endpoint" do
     headers = { "CONTENT_TYPE" => "application/json" }
-    post(api_v0x0_endpoints_url, :params => {:host => "example.com", :source_id => source.id.to_s, :tenant_id => tenant.id.to_s}.to_json)
```
bdunne: These actually need to be strings because JSON can't handle bigints I think

bdunne:
https://github.com/ManageIQ/topological_inventory-api/pull/48 fixes the issue

spec/controllers/api/v0x0/endpoints_controller_spec.rb
Fryguy: That's correct...we intentionally have ids as strings in the outgoing payload.

PR #51, opened 2018-11-19
Add API support for VMs and OrchestrationStacks

agrare:
This adds swagger definitions, controllers, and routes for VMs and OrchestrationStacks.

Depends on: https://github.com/ManageIQ/topological_inventory-core/pull/55


spec/swagger_spec.rb
```diff
@@ -1,3 +1,4 @@
+
```
bdunne: Extra newline?

agrare: :+1: removed, deleted a byebug require but forgot the extra line

PR #52, opened 2018-11-21
Update the .dockerignore

bdunne:


.dockerignore
```diff
@@ -1,3 +1,19 @@
...
+bundler.d
+Gemfile.lock
+v2_key
+/.bundle/
+/.rspec_status
+/.ruby-gemset
+/.ruby-version
+/.yardoc
+/_yardoc/
+/bundler.d/*
```
carbonin: Do we need this if we already have `bundler.d` above?

PR #56, opened 2018-11-30
Add service_plans/{id}/order to swagger.yaml docs

eclarizio:
An addendum to #37, this adds the definition for the order endpoint in the swagger.yaml.

@agrare Since Brandon is out at the moment, can you take a look and review? I think the only thing that may need changing are some of the descriptions, I don't know enough about the parameters to give a great concise and meaningful one so I feel like they're kind of redundant at the moment, but maybe the names of the properties is kind of self explanatory anyway.

### DIFF
--- a/public/doc/swagger-2-v0.0.1.yaml
+++ b/public/doc/swagger-2-v0.0.1.yaml
@@ -534,6 +534,52 @@ paths:
             "$ref": "#/definitions/ServicePlan"
         404:
           description: Not found
+  "/service_plans/{id}/order":
+    post:
+      summary: Order an existing ServicePlan
+      operationId: orderServicePlan
+      description: Returns a Task id
+      consumes:
+      - application/json
+      parameters:
+      - name: id
+        in: path
+        description: ID of the ServicePlan to order
+        required: true
+        type: string
+        pattern: '\A\d+\z'
+      - name: parameters
+        in: body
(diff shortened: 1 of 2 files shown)


public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -534,6 +534,52 @@ paths:
...
+                  type: string
+                namespace:
+                  type: string
+            provider_control_parameters:
+              description: Extra parameters defining the provider control
+              type: object
+              properties:
+                namespace:
+                  type: string
+                OpenShift_param1:
```
agrare: @eclarizio what is OpenShift_param1 used for?  I don't see it referenced by the core code for ordering.

agrare:
Just the one question but since the parameters match what the controller expects we can address in a followup to delete both if it isn't used.

public/doc/swagger-2-v0.0.1.yaml
eclarizio: @agrare Good question, I have no idea, I was just following the parameter list from the original code I had moved over. ¯\_(ツ)_/¯

agrare: Yeah understood, @mkanoor do you know what this parameter is and if it is safe to remove?

PR #57, opened 2018-12-03
Remove extranneous properties from swagger yaml and update descriptions

eclarizio:
As @mkanoor suggested in #56, this PR removes the extra properties that were defined on `service_parameters` and `provider_control_parameters` so that any hash that gets passed to the endpoint is acceptable.

@agrare Please review/merge when ready 😄


mkanoor:
👍

agrare:
@eclarizio @mkanoor can we remove `:OpenShift_param1` from https://github.com/ManageIQ/topological_inventory-api/blob/354cbf747d9da9328828f9da4e945287ff044d26/app/controllers/api/v0/service_plans_controller.rb#L28 also?

You don't need these in the swagger doc but you do need to permit them in the controller still right?  I thought you had to permit each key in the hash or rails rejected the payload.

eclarizio:
@agrare Yeah, that's what I found when I was testing it. There's no way that I know of to simply permit a nested hash just by permitting the parent key. If that's the case though, we would need to know all of the potential parameter key names that can be passed through, right?

mkanoor:
@agrare @eclarizio 
We have to 2 keys service_parameters and provider_control_parameters
For service_parameters there is a JSON schema that lists all the parameter names, constraints and required ones. We fetch this JSON schema as part of the service plans. The UI fills this and it we send it as a hash under service_parameters.

The provider_control_parameters is our creation that allows us to define parameters needed for the provider or source. On OpenShift we need the name of the namespace where the new service should be provisioned. We had thought of having a JSON schema per provider that we could fetch and have the user fill-in the provider specific parameters.

When we add these 2 parameters into our tables we are not validating these 2 fields
https://github.com/ManageIQ/service_portal-api/blob/6ff5f27cdd0a0c031fa823b4a6f727d1e60f6e9a/app/services/add_to_order.rb#L14

agrare:
So how about we use the json schema to get the list of permitted keys?

mkanoor:
@agrare 
That will change from service to service, if you are installing MARIA_DB you have one set of keys, if you are installing NGINX you have another set of keys

mkanoor:
@agrare Where is a hardcoded list, we are using hashes for service parameters which is fed from a JSON schema.
For the provider control parameters that hash would have to be sent to the provider specific API so it would know what to use from the hash.

agrare:
@mkanoor https://github.com/ManageIQ/topological_inventory-api/blob/354cbf747d9da9328828f9da4e945287ff044d26/app/controllers/api/v0/service_plans_controller.rb#L26-L29

mkanoor:
That needs to be empty hashes or you would have to fetch the schema for the service plan and do an additional validation at runtime.

agrare:
Yup that's exactly what I was saying [here](https://github.com/ManageIQ/topological_inventory-api/pull/57#issuecomment-443868414) :laughing:

mkanoor:
@agrare @eclarizio 
The service_parameters and provider_control_parameters are a bit like the options hash that we store per request in our database for ManageIQ. In ManageIQ we do the validation before we create the record making it a synchronous call which blocks the worker and has the caller waiting for getting a request. In the past @Fryguy has suggested that to reduce this bottleneck we should first create a request and send the request id/url back to the caller so it can continue and in the background when the Request is processed the first thing it does is the validation and updates the request to indicate if their are any errors in the process. Even @eclarizio is familiar with the service dialog validation that happens in ManageIQ when a provision request comes in and doing it in line has all kinds of issues. The validation of the service parameters or provider control parameters could happen later after the record is created and we should not do it as part of the strong parameter checks which happens before the record is created. The Service Dialog in ManageIQ is similar to the JSON schema that we are seeing for plans. The UI does the validation before it sends the parameters, then it comes to the service portal which has to do the validation because that data needs to be sent to the approval service and then it comes to the topology service, so we already have multiple places where the validation is happening adding a third one inline and not delaying it till the request processing starts will make the process slower. @Fryguy comments?

PR #58, opened 2018-12-05
Set the Source.uid in API spec tests

agrare:
https://github.com/ManageIQ/topological_inventory-core/pull/74 made the source uid mandatory so update the API specs accordingly.

### DIFF
--- a/app/controllers/api/v0/sources_controller.rb
+++ b/app/controllers/api/v0/sources_controller.rb
@@ -7,7 +7,7 @@ module Api
       include Api::Mixins::UpdateMixin
 
       def create
-        source = Source.create!(create_params)
+        source = Source.create!(create_params.merge!("uid" => SecureRandom.uuid))
         render :json => source, :status => :created, :location => api_v0x0_source_url(source.id)
       end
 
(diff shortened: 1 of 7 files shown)


app/controllers/api/v0/sources_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
       private
 
       def create_params
-        body_params.permit(:name, :source_type_id, :tenant_id)
+        body_params.permit(:name, :source_type_id, :tenant_id, :uid)
```
carbonin: Should we really be having the client give us a uid? Would it be better to generate one on the model?

agrare: I was thinking there might be cases where we're using the native provider uuid but that doesn't apply to any of these providers yet so yeah i'll just generate it

carbonin: Does this line still need to be here? Or can this be removed? It looks like we'll overwrite anything that is passed in the params based on the `.merge!` above, right?

agrare: Yup just forgot to delete it

bdunne:
@agrare I don't think this should be here, is there a follow-up to the core repo to set this with `default_value_for` or something else on the model?

PR #64, opened 2018-12-10
Use File.join for building the route prefix and deal with extra slashes

carbonin:


config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   prefix = "api"
-  prefix = "#{ENV["PATH_PREFIX"]}/#{ENV["APP_NAME"]}" if ENV["PATH_PREFIX"].present? && ENV["APP_NAME"].present?
+  if ENV["PATH_PREFIX"].present? && ENV["APP_NAME"].present?
+    prefix = File.join(ENV["PATH_PREFIX"], ENV["APP_NAME"]).gsub(/^\/+|\/+$/, "")
```
bdunne: Does the regex need to be that complex?

```ruby
irb(main):003:0> File.join("//example/path/prefix/", "////appname/").gsub(/\/+/,"/")
=> "/example/path/prefix/appname/"
irb(main):004:0> File.join("//example/path/prefix/", "/appname/").gsub(/\/+/,"/")
=> "/example/path/prefix/appname/"
```

PR #65, opened 2018-12-11
Create Task for service ordering and update its status synchronously

eclarizio:
The task creation is being moved from the core repo and the `ServicePlan#order` method to here, so that we can then publish a message to the messaging service later with the `task` and the `service_plan`.

Basically `#order_service_plan_and_update_task` is what I think the subscriber of the messages needs to do, so I separated it into a small private method.

Note that this needs [this PR from the core repo]() since the `#order` method is now expected to return the task context and not a task id.


spec/controllers/api/v0x0/service_plans_controller_spec.rb
```diff
@@ -33,20 +33,29 @@
...
+        @body = JSON.parse(response.body)
+      end
+
+      it "returns json with the task id" do
+        expect(@body).to have_key("task_id")
+      end
+
+      it "sets the context based on the results of the order and sets the status of the task to completed" do
+        task = Task.find(@body["task_id"])
+        expect(task.context).to eq("task_context")
```
bdunne: `expect(Task.find(@body["task_id"])).to have_attributes…`?

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -6,14 +6,25 @@ class ServicePlansController < ApplicationController
...
+        render :json => {:task_id => task.id}
       rescue ActiveRecord::RecordNotFound
         head :bad_request
       end
 
       private
 
+      def order_service_plan_and_update_task(task, service_plan)
+        task.context = service_plan.order(order_params)
+        task.status = "completed"
```
mkanoor: @eclarizio In my testing I have seen that the service_plan.order sends a request to OCP and comes back right away without waiting for the provision to end, so is it appropriate to mark the status as completed here. Unless its just a place holder.

agrare: @mkanoor this is out of scope of this change

PR #72, opened 2018-12-12
Hardcode the basePath in the swagger yml for client generation

Fryguy:
@carbonin Please review.


Fryguy:
cc @mkanoor @bzwei I think we will need to do this to service_portal and approval as well.  Additionally, I think both service_portal and approval need to rename the file to swagger-2-v0.0.1.yaml

Fryguy:
@bdunne Also please review.

PR #74, opened 2018-12-13
Use pagination on collection responses

bdunne:
Based on https://jsonapi.org/ and https://jsonapi.org/format/#fetching-pagination which I found somewhere in Swagger / OpenAPI documentation.  The IPP for the API doesn't mention how to implement pagination

https://swagger.io/docs/specification/describing-parameters/ and https://swagger.io/docs/specification/2-0/describing-parameters/ refer to `limit` and `offset`, but I wouldn't mind changing to `size` and `page_number` or something like that if people have strong opinions.


Fryguy:
@bdunne Is this based on some pattern or some known thing that's documented out there? I like it, but just wanted some context.

bdunne:
@Fryguy I updated the description with some links.  Either `limit` & `offset` or `page_size` & `page_number` are fine with me.  I'm going to refactor this into something more reusable, but it was an initial spike and I like the way that the structure turned out.

I think @martinpovolny was also talking about JSON:API with `limit` & `offset`.  Any thoughts on this approach Martin?

martinpovolny:
Unfortunately the section on pagination in the IPP is not very detailed.

> Each paginated response should include metadata describing pagination parameters, including total number of records, total number of pages, and hypermedia links to the first, last, next, and previous pages of results.

This is what is important from my perspective. The metainformation should be there so that the client does not have to fire extra requests just to figure the number of records or default paging settings.

I personally prefer offset/limit as it is easy to work with and it nicely maps to the underlying SQL database. But yes, it was not included in the IPP.

To work effectively I'd like to have the number of pages, the offset and the limit in all responses. Also I'd like to see search/filtering options that where used to generate a response (not relevant now, I know). There was a nice comment on this by the Satellite folks in the IPP but it also did not get included in the final IPP.

martinpovolny:
ping @Fryguy ?

bdunne:
This is marked as wip because I'm trying to get it into https://github.com/ManageIQ/manageiq-api-common/pull/5.

bdunne:
@Fryguy Okay, this PR is updated to use the paginated response code from manageiq-api-common.

bdunne:
FYI... This will break existing consumers of the api since the collection response is different. cc @eclarizio @gmcculloug @mkanoor

gmcculloug:
cc @lgalis @syncrou

Fryguy:
@bdunne I am :+1: on this PR, but I want to make sure that all of the clients are in place so we don't have any surprises for the callers.  As discussed, we have both JavaScript and Ruby clients at a minimum that should be ready before merge.

bdunne:
Okay, I re-worked the PR to add pagination but only to version 0.1+.  So v0.0 is unchanged allowing existing callers to continue as they were before.  I also left the /v0 redirect pointing to v0.0 until we have confirmation that the major users have migrated to v0.1 clients.

bdunne:
@Fryguy I think this is ready, any other concerns?

PR #76, opened 2018-12-19
Workaround change to needs_migration? in rails 5.2

agrare:
In rails 5.2.0 there was a change to
ActiveRecord::Migrator.needs_migration? and it was moved under a
MigrationContext class.

Introduced by https://github.com/rails/rails/pull/31727

### DIFF
--- a/lib/tasks/deployment.rake
+++ b/lib/tasks/deployment.rake
@@ -1,7 +1,7 @@
 module ApplicationDeployment
   def self.status
     return "new_deployment" if ActiveRecord::Migrator.current_version.zero?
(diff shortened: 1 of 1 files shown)


agrare:
cc @jrafanie @carbonin

carbonin:
Opened https://github.com/ManageIQ/topological_inventory-api/pull/77 as an alternative. I don't think we actually need this logic.

agrare:
:+1: works for me

jrafanie:
👍 thanks for the headup

jrafanie:
Or headsup

PR #78, opened 2018-12-19
Fix issues with swagger definition and tests

bdunne:
- Alphabetize paths and definitions for readability
- Fix whitespace
- Properly escape regex
- Fix quoting issues
- Fix typo `readonly` to `readOnly`
- Describing an attribute / parameter with `$ref` and local definitions needs to be wrapped in `allOf:`
- Load swagger regexes as real regexes for validation purposes
- Revert changes to optional _id attributes
- Fix unknown reference issues during validation in rspec matcher

### DIFF
--- a/lib/open_api/docs/doc_v2.rb
+++ b/lib/open_api/docs/doc_v2.rb
@@ -12,10 +12,11 @@ module OpenApi
       end
 
       def definitions
-        @definitions ||= content["definitions"]
+        @definitions ||= substitute_regexes(content["definitions"])
+      end
 
       def parameters
-        @parameters ||= content["parameters"]
+        @parameters ||= substitute_regexes(content["parameters"])
       end
 
       def expanded_definition(key)
@@ -65,6 +66,23 @@ module OpenApi
       def fetch_ref_value(ref_path)
         _, section, property = ref_path.split("/")
         send(section)[property]
+      end
+
+      def substitute_regexes(object)
+        if object.kind_of?(Array)
+          object.collect { |i| substitute_regexes(i) }
+        elsif object.kind_of?(Hash)
+          object.each do |k, v|
+            object[k] = k == "pattern" ? regexp_from_pattern(v) : substitute_regexes(v)
+          end
+        else
+          object
+        end
+      end
+
+      def regexp_from_pattern(pattern)
+        raise "Pattern #{pattern.inspect} is not a regular expression" unless pattern.starts_with?("/") && pattern.ends_with?("/")
+        Regexp.new(pattern[1..-2])
       end
     end
   end
--- a/lib/open_api/docs/doc_v2.rb
+++ b/lib/open_api/docs/doc_v2.rb
@@ -13,6 +13,15 @@ module OpenApi
 
       def definitions
         @definitions ||= content["definitions"]
+
+      def parameters
+        @parameters ||= content["parameters"]
+      end
+
+      def expanded_definition(key)
+        definitions[key].tap do |i|
+          i["properties"] = substitute_references(i["properties"])
+        end
       end
 
       def example_attributes(key)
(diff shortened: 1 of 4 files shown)


lib/open_api/docs/doc_v2.rb
```diff
@@ -39,6 +51,44 @@ def routes
...
+      private
+
+      def substitute_references(object)
+        if object.kind_of?(Array)
+          object.collect { |i| substitute_references(i) }
+        elsif object.kind_of?(Hash)
+          return fetch_ref_value(object["$ref"]) if object.keys == ["$ref"]
+          object.each_with_object({}) do |(k, v), h|
+            h[k] = substitute_references(v)
+          end
```
Fryguy: I don't think you need to create a new Hash here...can this work?

```ruby
          return fetch_ref_value(object["$ref"]) if object.keys == ["$ref"]
          object.keys.each do |k|
            object[k] = substitute_references(object[k])
          end
```

lib/open_api/docs/doc_v2.rb
```diff
@@ -39,6 +51,44 @@ def routes
...
+        send(section)[property]
+      end
+
+      def substitute_regexes(object)
+        if object.kind_of?(Array)
+          object.collect { |i| substitute_regexes(i) }
+        elsif object.kind_of?(Hash)
+          object.each_with_object({}) do |(k, v), h|
+            h[k] = k == "pattern" ? regexp_from_pattern(v) : substitute_regexes(v)
+          end
```
Fryguy: same here...can this work?

```ruby
object.keys.each do |k|
  v = object[k]
  object[k] = k == "pattern" ? regexp_from_pattern(v) : substitute_regexes(v)
end
```

lib/open_api/docs/doc_v2.rb
Fryguy: Note I'm doing .keys.each here to avoid modifying a hash while it's being iterated over

lib/open_api/docs/doc_v2.rb
```diff
@@ -39,6 +51,44 @@ def routes
...
+        else
+          object
+        end
+      end
+
+      def regexp_from_pattern(pattern)
+        match = pattern.match(/\A\/(.*)\/\z/)
```
Fryguy: The \A feels like it's checked wrong...I think the regex is supposed to be  `/\\\A\/(.*)\/\z/`.  You may also want to anchor it with actual \A and \z

lib/open_api/docs/doc_v2.rb
```diff
@@ -12,7 +12,19 @@ def version
...
+        @definitions ||= substitute_regexes(content["definitions"])
+      end
+
+      def parameters
+        @parameters ||= substitute_regexes(content["parameters"])
+      end
+
+      def expanded_definition(key)
+        definitions[key].tap do |i|
+          i["properties"] = i["properties"].each_with_object({}) do |(k, v), hash|
```
Fryguy: This feels like it could be

```ruby
i["properties"] = substitute_references(i["properties"])
```

and then it can just reuse the Hash logic below.

lib/open_api/docs/doc_v2.rb
bdunne: Sorry, I don't understand the use case of the extra `\\`.  Wouldn't that try to match a `\` before the start of the line?

I'm expecting the input `pattern` to be something like `/^d+$/` and in that case I want to capture only the `^d+$`.  So I was looking for start of string, then `/`, then capture everything until the next `/`, then the end of the string.

Fryguy: As discussed, if we assume a pattern will always be a regex and that it always must start with / and end with /, then let's just check that and then blow up if it's wrong.  Perfer start_with and end_woth aso, for readability.

PR #79, opened 2018-12-20
Add schema to params for SourceTypes.

martinpovolny:
@agrare, @bdunne  : Do I need this to be able to access the `:schema` via the API?

### DIFF
--- a/app/controllers/api/v0/source_types_controller.rb
+++ b/app/controllers/api/v0/source_types_controller.rb
@@ -12,7 +12,7 @@ module Api
       private
 
       def create_params
-        body_params.permit(:name, :product_name, :vendor)
+        body_params.permit(:name, :product_name, :vendor, :schema)
       end
 
       def list_params
--- a/public/doc/swagger-2-v0.0.1.yaml
+++ b/public/doc/swagger-2-v0.0.1.yaml
@@ -1655,6 +1655,8 @@ definitions:
       vendor:
         type: string
         example: Red Hat
+      schema:
+        type: object
   Task:
     type: object
     properties:


app/controllers/api/v0/source_types_controller.rb
```diff
@@ -12,11 +12,11 @@ def create
       private
 
       def create_params
-        body_params.permit(:name, :product_name, :vendor)
+        body_params.permit(:name, :product_name, :vendor, :schema)
```
bdunne: :+1:

app/controllers/api/v0/source_types_controller.rb
```diff
@@ -12,11 +12,11 @@ def create
...
       end
 
       def list_params
-        params.permit(:name, :product_name, :vendor)
+        params.permit(:name, :product_name, :vendor, :schema)
```
bdunne: I don't think we'll need `:schema` here.  This is the params that can be passed to `model.where`, and I don't think people will search / filter by `:schema`.  I should probably come up with a better name for this method.  :smile:

agrare:
> Do I need this to be able to access the :schema via the API?

For that I think you just need to add it to the [swagger yaml](https://github.com/ManageIQ/topological_inventory-api/blob/master/public/doc/swagger-2-v0.0.1.yaml#L1638-L1657)

app/controllers/api/v0/source_types_controller.rb
martinpovolny: fixing it...

martinpovolny:
> For that I think you just need to add it to the swagger yaml

and run some command, or just add it?

martinpovolny:
That is can I close this PR or do I actually need to do something?

agrare:
> and run some command, or just add it?

Nope just add it, the serializer mixin looks to the swagger yaml to return permitted keys: https://github.com/ManageIQ/topological_inventory-api/blob/master/config/initializers/application_record_serializing.rb#L7

> That is can I close this PR or do I actually need to do something?

I think you can just add it to this PR since it goes together with adding `:schema` to the create params

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -1655,6 +1655,8 @@ definitions:
       vendor:
         type: string
         example: Red Hat
+      schema:
+        type: string
```
agrare: @bdunne since schema is a jsonb type would `type: object` be more appropriate here or is string fine?

martinpovolny:
Ok. Removed the `list_param` change and hand-edited the swagger yaml.

public/doc/swagger-2-v0.0.1.yaml
bdunne: Yes, that's what I was thinking

agrare:
@martinpovolny if you change schema type to object in the swagger yaml this looks good to me

PR #80, opened 2018-12-20
Filter out encrypted values on serialization

bdunne:
Good practice, but also necessary preparation for Authentications endpoint

### DIFF
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -2,11 +2,12 @@ module OpenApi
   module Serializer
     def as_json(arg = {})
       previous = super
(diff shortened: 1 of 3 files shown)


lib/open_api/serializer.rb
```diff
@@ -2,11 +2,12 @@ module OpenApi
   module Serializer
     def as_json(arg = {})
       previous = super
-      return previous unless arg.key?(:prefixes)
+      encryption_filtered = previous.except(*self.class.try(:encrypted_columns))
```
Fryguy: you can replace `previous` with `super` directly here, as the variable doesn't help for clarity anymore.

PR #81, opened 2018-12-20
Add lives_on_type/id to ContainerNode

agrare:
I'd prefer to return a lives_on url rather than returning a type/id but was having issues getting this into the controller since it had already gone through `as_json`.

This will unblock the AI team and allow them to get started with their models /cc @bronaghs


bdunne:
Yeah, there are several places I think we should return a URL to reference other objects, but let's tackle that generically in a separate PR.

PR #89, opened 2019-01-10
Expose tagging

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/83
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/93

### DIFF
--- a/app/controllers/api/v0/container_images_controller.rb
+++ b/app/controllers/api/v0/container_images_controller.rb
@@ -4,10 +4,6 @@ module Api
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
 
-      def index
-        render json: model.where(list_params).includes(:container_image_tags => :tag).references(:container_image_tags => :tag)
-      end
-
       private
 
       def list_params
--- a/app/controllers/api/v0/container_images_controller.rb
+++ b/app/controllers/api/v0/container_images_controller.rb
@@ -4,6 +4,10 @@ module Api
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
 
+      def index
+        render json: model.where(list_params).includes(:container_image_tags => :tag).references(:container_image_tags => :tag)
+      end
+
       private
 
       def list_params
--- a/app/controllers/api/v0/mixins/index_mixin.rb
+++ b/app/controllers/api/v0/mixins/index_mixin.rb
@@ -3,7 +3,17 @@ module Api
     module Mixins
       module IndexMixin
         def index
-          render json: model.where(list_params)
+          render json: scoped(model.where(list_params))
+        end
+
+        def scoped(relation)
+          if model.respond_to?(:taggable?) && model.taggable?
+            ref_schema = {model.tagging_relation_name => :tag}
+
+            relation = relation.includes(ref_schema).references(ref_schema)
+          end
+
+          relation
         end
       end
     end
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -11,6 +11,7 @@ module OpenApi
       schema["properties"].keys.each do |name|
         attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
         attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
+        attrs[name] = self.public_send(name) unless attrs.key?(name)
       end
       attrs.compact
     end
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -2,13 +2,13 @@ module OpenApi
   module Serializer
     def as_json(arg = {})
       previous = super
-      encryption_filtered = previous.except(*self.class.try(:encrypted_columns))
+      encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
+      encryption_filtered = previous.except(*encrypted_columns_set)
       return encryption_filtered unless arg.key?(:prefixes)
       /\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ arg[:prefixes].first
       version = [major, minor].compact.join(".")
       schema  = Api::Docs[version].definitions[self.class.name]
       attrs   = encryption_filtered.slice(*schema["properties"].keys)
-      encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
 
       schema["properties"].keys.each do |name|
         attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -8,10 +8,12 @@ module OpenApi
       version = [major, minor].compact.join(".")
       schema  = Api::Docs[version].definitions[self.class.name]
       attrs   = encryption_filtered.slice(*schema["properties"].keys)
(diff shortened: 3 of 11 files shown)


Ladas:
cc @bdunne @gtanzillo @agrare so this is the draft and it needs https://github.com/ManageIQ/topological_inventory-core/pull/83/commits/c5a91019ee50f81cca7ec681fc176cd375a9d0d0

Then the API response for container_images looks like:
```
{
        "id": "193",
        "tenant_id": "10",
        "source_id": "4",
        "name": "jboss-datavirt-6/datavirt63-openshift",
        "source_created_at": "2018-12-12T09:16:03Z",
        "taggings": [
            {
                "tag_id": 1,
                "name": "architecture",
                "value": "x86_64"
            },
            {
                "tag_id": 2,
                "name": "authoritative-source-url",
                "value": "registry.access.redhat.com"
            },
            {
                "tag_id": 3,
                "name": "build-date",
                "value": "2017-04-21T09:07:49.952292"
            },

```

lib/open_api/serializer.rb
```diff
@@ -11,6 +11,7 @@ def as_json(arg = {})
       schema["properties"].keys.each do |name|
         attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
         attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
+        attrs[name] = self.public_send(name) unless attrs.key?(name)
```
Ladas: looks like this would clash with encryption_filtered

Ladas: @bdunne btw. why are we turning id attributes into strings? Those should remain integers? Also in the swagger yaml, we are saying these are format: uuid, but those are integers.

app/controllers/api/v0/container_images_controller.rb
```diff
@@ -4,6 +4,10 @@ class ContainerImagesController < ApplicationController
       include Api::Mixins::IndexMixin
       include Api::Mixins::ShowMixin
 
+      def index
+        render json: model.where(list_params).includes(:container_image_tags => :tag).references(:container_image_tags => :tag)
```
Ladas: this will make sure we do only 1 query, then the model touching these takes it from the cache

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -1796,6 +1801,40 @@ definitions:
         example: Red Hat
       schema:
         type: object
+  Tag:
+    type: object
+    properties:
+      tag_id:
+        type: string
+        readOnly: true
+        format: uuid
```
Ladas: @bdunne should be integer? Or any kin of validation we would do will fail on this

app/controllers/api/v0/container_images_controller.rb
Ladas: @bdunne it should not be super hard to generate this automatically upon some condition, like model.taggable?

public/doc/swagger-2-v0.0.1.yaml
bdunne: See https://github.com/ManageIQ/topological_inventory-api/commit/3bc29df178e71c7174ed2c3299ec0de0d389cb15#diff-93e2df6f7892ddefd3d8b61839de743fR1412

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -1389,6 +1389,11 @@ definitions:
         type: string
         format: date-time
         readOnly: true
+      taggings:
+        type: array
+        items:
+          "$ref": "#/definitions/Tagging"
```
Ladas: @bdunne hm, the specs require this to be a class. Should I create a class for this, just and instantiate it in https://github.com/ManageIQ/topological_inventory-core/commit/c5a91019ee50f81cca7ec681fc176cd375a9d0d0 ?

Or what is the right way? I assume we want to have this type check, that it can't be just array of anything.

bdunne: You can add it to the exception list for now https://github.com/ManageIQ/topological_inventory-api/blob/master/spec/swagger_spec.rb#L105
I'll probably have to do something different as we accumulate more non-model definitions.

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -1358,6 +1358,11 @@ definitions:
         type: string
         readOnly: true
         format: uuid
+      taggings:
```
Ladas: @agrare @bdunne do we prefer ```taggings``` or ```tag_list``` or something else? Relation is already named tags, so we can't use that.

bdunne: I guess taggings.

The more we talk about this, the more I think tags should be a subcollection

Ladas: Can we do subcollections so we don't have to do N+1 queries to retrieve VMs with tags?

bdunne: Not that I'm aware of

Ladas: Hm so subcollection will be good that we can paginate it, in a case that there's too much tags per entity (or we could limit it in this query maybe). But otherwise the usability will not be that good.

I think it'd be a good idea to start to design how the tag interface will look like (mainly filtering by tags). Then it could make it more clear what his should look like.

bdunne: We would not be able to paginate or limit this as an attribute, which we would be able to do with a subcollection.

As far as performance goes, we're limited by OpenApi / Swagger.

Either way, I think we should discuss the use cases first. cc @gtanzillo

Ladas:
@bdunne can you review, please? This is what we've agreed should be ok for now.

lib/open_api/serializer.rb
```diff
@@ -8,9 +8,12 @@ def as_json(arg = {})
       version = [major, minor].compact.join(".")
       schema  = Api::Docs[version].definitions[self.class.name]
       attrs   = encryption_filtered.slice(*schema["properties"].keys)
+      encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
```
agrare: Could you do this up [here](https://github.com/ManageIQ/topological_inventory-api/blob/cfef5d522435656287ebe3b8389606660508ea89/lib/open_api/serializer.rb#L5) and reuse the encrypted_columns_set for the encryption_filtered?

Ladas: done

Ladas:
@bdunne @agrare can you review please? I've added the collections also to v0.1 since that was just added

app/controllers/api/v0/mixins/index_mixin.rb
```diff
@@ -3,7 +3,17 @@ module V0
...
-          render json: model.where(list_params)
+          render json: scoped(model.where(list_params))
+        end
+
+        def scoped(relation)
+          if model.respond_to?(:taggable?) && model.taggable?
+            ref_schema = {model.tagging_relation_name => :tag}
+
+            relation = relation.includes(ref_schema).references(ref_schema)
+          end
```
agrare: Minor but I wonder if we can move this to the tagging mixin in core?

Ladas: I guess that could work

agrare: Would be a good follow-up PR, then we can drop this from the two mixins here

PR #93, opened 2019-01-17
Use the manageiq-api-common gem for adding metrics middleware

carbonin:
Requires https://github.com/ManageIQ/manageiq-api-common/pull/23

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,9 +16,6 @@ require "rails/test_unit/railtie"
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-require 'prometheus/middleware/collector'
-require 'prometheus/middleware/exporter'
-
 module TopologicalInventory
   class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -39,7 +36,7 @@ class Application < Rails::Application
                       ManageIQ::Loggers::Base.new(Rails.root.join("log", "#{Rails.env}.log"))
                     end
 
-    config.middleware.use(Prometheus::Middleware::Collector)
-    config.middleware.use(Prometheus::Middleware::Exporter)
+    require 'manageiq-api-common'
```
bdunne: Do you need this require?  Also, any reason to be marked as WIP?

carbonin: Nope, and maybe.

I want to update common to accept a metrics prefix. Right now all the metrics are called `http_server_<metric>` *for all of the apps* So I think it would be better to update that first then allow this merge to deploy the change.

carbonin:
Now also requires https://github.com/ManageIQ/manageiq-api-common/pull/26

config/application.rb
```diff
@@ -39,7 +36,6 @@ class Application < Rails::Application
                       ManageIQ::Loggers::Base.new(Rails.root.join("log", "#{Rails.env}.log"))
                     end
 
-    config.middleware.use(Prometheus::Middleware::Collector)
-    config.middleware.use(Prometheus::Middleware::Exporter)
+    ManageIQ::API::Common::Metrics.activate(config, "topological_inventory_api")
```
bdunne: Should it be `topological_inventory-api`

bdunne: Okay

PR #99, opened 2019-01-22
Disable HTTP PUT

bdunne:
Rails sends PUT and PATCH to :update, they're not exactly the same and I don't think we need to "replace" records as PUT suggests.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,11 +1,14 @@
 Rails.application.routes.draw do
+  # Disable PUT for now since rails sends these :update and they aren't really the same thing.
+  def put(*_args); end
+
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -1,11 +1,14 @@
 Rails.application.routes.draw do
+  # Disable PUT for now since rails sends these :update and they aren't really the same thing.
+  def put(*args, &block); end
```
Fryguy: Minor, but per rubocop style, these should be `*_args, &_block`, or even better, you don't even need to capture the block so...

```ruby
def put(*); end
```

PR #102, opened 2019-01-25
Switch to autoload in production

bdunne:
Started GET \"/r/insights/platform/topological-inventory/v0.1/sources\" for 52.36.139.101 at 2019-01-25 14:12:41 +0000
Processing by Api::V0x1::SourcesController#index as */*
Completed 500 Internal Server Error in 4ms (ActiveRecord: 0.0ms)
NameError (uninitialized constant ManageIQ::API::Common::PaginatedResponse):
app/controllers/api/v0x1/mixins/index_mixin.rb:6:in `index'

Eagerload is disabling autoload causing errors in production that are not seen in development.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -4,11 +4,14 @@ Rails.application.configure do
   # Code is not reloaded between requests.
   config.cache_classes = true
 
-  # Eager load code on boot. This eager loads most of Rails and
+  # Don't eager load code on boot. This eager loads most of Rails and
   # your application in memory, allowing both threaded web servers
   # and those relying on copy on write to perform better.
   # Rake tasks automatically ignore this option for performance.
-  config.eager_load = true
+  # Enabling eagerloading also disables autoloading which causes problems
+  # in production that don't happen in development.
+  config.eager_load_paths = []
+  config.eager_load = false
 
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local       = false
--- a/config/application.rb
+++ b/config/application.rb
@@ -26,7 +26,21 @@ module TopologicalInventory
     # Middleware like session, flash, cookies can be added back manually.
     # Skip views, helpers and assets when generating a new resource.
     config.api_only = true
-    config.autoload_paths << Rails.root.join('lib')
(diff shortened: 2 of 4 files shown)


Fryguy:
I don't understand how this could happen in a clean room build.  Only thing I can think of is that the docker build is not set to nocache

carbonin:
@bdunne where were you seeing this? We set the build to `--no-cache` in the ci environment for exactly this reason, but we still don't trigger builds on pushes to core/common. That's why I opened https://projects.engineering.redhat.com/browse/TPINVTRY-147

Dockerfile
```diff
@@ -18,21 +18,20 @@ ENV WORKDIR /opt/topological_inventory-api/
 ENV RAILS_ROOT $WORKDIR
 WORKDIR $WORKDIR
 
-COPY Gemfile $WORKDIR
+COPY . $WORKDIR
+COPY docker-assets/entrypoint /usr/bin
+
+RUN chgrp -R 0 $WORKDIR && \
+    chmod -R g=u $WORKDIR
```
carbonin: Why move this up? Could this move cause an issue if the gems are installed in `$WORKDIR`?

carbonin:
Did this PR change from editing the Dockerfile to changing autoload? That makes a bunch of these comments a bit odd in retrospect.

bdunne:
> Did this PR change from editing the Dockerfile to changing autoload?

Yes, it did.

config/environments/production.rb
```diff
@@ -4,11 +4,13 @@
   # Code is not reloaded between requests.
   config.cache_classes = true
 
-  # Eager load code on boot. This eager loads most of Rails and
+  # Don't eager load code on boot. This eager loads most of Rails and
   # your application in memory, allowing both threaded web servers
   # and those relying on copy on write to perform better.
   # Rake tasks automatically ignore this option for performance.
-  config.eager_load = true
+  # Eagerloading also disables autoloading :-( causing problems in production that don't happen in development
```
Fryguy: "Disabling eager loading also disables autoloading ..."

PR #103, opened 2019-01-25
Async service ordering

eclarizio:
This will be used in combination with the operations openshift messaging service to handle the published messages.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,14 @@
 source 'https://rubygems.org'
 
-gem 'jbuilder',          '~> 2.0'
-gem 'json-schema',       '~> 2.8'
-gem 'manageiq-loggers',  '~> 0.1'
-gem 'manageiq-password', '~> 0.2', ">= 0.2.1"
-gem 'pg',                '~> 1.0', :require => false
-gem 'puma',              '~> 3.0'
-gem 'rack-cors',         '>= 0.4.1'
-gem 'rails',             '~> 5.2.2'
+gem 'jbuilder',           '~> 2.0'
+gem 'json-schema',        '~> 2.8'
+gem 'manageiq-loggers',   '~> 0.1'
+gem 'manageiq-messaging', '~> 0.1.2'
+gem 'manageiq-password',  '~> 0.2', ">= 0.2.1"
+gem 'pg',                 '~> 1.0', :require => false
+gem 'puma',               '~> 3.0'
+gem 'rack-cors',          '>= 0.4.1'
+gem 'rails',              '~> 5.2.2'
 
 gem 'inventory_refresh', :git => 'https://github.com/ManageIQ/inventory_refresh', :branch => 'master'
 gem 'manageiq-api-common', :git => 'https://github.com/ManageIQ/manageiq-api-common', :branch => 'master'
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'jbuilder',           '~> 2.0'
 gem 'json-schema',        '~> 2.8'
 gem 'manageiq-loggers',   '~> 0.1'
-gem 'manageiq-messaging', '~> 0.1.2'
+gem 'manageiq-messaging', '~> 0.1.2', :require => false
 gem 'manageiq-password',  '~> 0.2', ">= 0.2.1"
 gem 'pg',                 '~> 1.0', :require => false
 gem 'puma',               '~> 3.0'
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -10,7 +10,7 @@ module Api
 
         messaging_client.publish_message(
           :service => "platform.topological-inventory.operations-openshift",
-          :message => "order_service",
+          :message => "ServicePlan.order",
           :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => params_for_order}
         )
 
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -5,13 +5,13 @@ module Api
       include Api::V0::Mixins::ShowMixin
 
       def order
-        service_plan = model.find(order_params[:service_plan_id].to_i)
+        service_plan = model.find(params_for_order[:service_plan_id].to_i)
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
         messaging_client.publish_message(
           :service => "platform.topological_inventory.operations-openshift",
           :message => "order_service",
-          :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => order_params}
+          :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => params_for_order}
         )
 
         render :json => {:task_id => task.id}
@@ -25,7 +25,7 @@ module Api
         params[:service_plan_id].to_i
       end
 
-      def order_params
+      def params_for_order
         params.permit(
           :service_plan_id,
           :service_parameters          => {},
@@ -33,7 +33,7 @@ module Api
         ).to_h
       end
 
-      def list_params
+      def params_for_list
         params.permit(:source_id, :tenant_id, :service_offering_id)
       end
 
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -11,7 +11,7 @@ module Api
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
         messaging_client.publish_message(
-          :service => "topological_inventory-orderer",
+          :service => "platform.topological_inventory.openshift-operations",
           :message => "order_service",
           :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => order_params}
         )
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -11,7 +11,7 @@ module Api
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
         messaging_client.publish_message(
-          :service => "platform.topological_inventory.openshift-operations",
+          :service => "platform.topological_inventory.operations-openshift",
           :message => "order_service",
           :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => order_params}
         )
--- a/app/controllers/api/v0/service_plans_controller.rb
+++ b/app/controllers/api/v0/service_plans_controller.rb
@@ -9,7 +9,7 @@ module Api
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
         messaging_client.publish_message(
-          :service => "platform.topological_inventory.operations-openshift",
+          :service => "platform.topological-inventory.operations-openshift",
           :message => "order_service",
           :payload => {:task_id => task.id, :service_plan_id => service_plan.id, :order_params => params_for_order}
         )
@@ -50,10 +50,6 @@ module Api
           :port     => ENV["QUEUE_PORT"] || "9092",
           :encoding => "json"
         }
-      end
-
-      def model
-        ServicePlan
       end
     end
   end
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -1,13 +1,15 @@
...
-gem 'puma',              '~> 3.0'
-gem 'rack-cors',         '>= 0.4.1'
-gem 'rails',             '~> 5.2.2'
+gem 'jbuilder',           '~> 2.0'
+gem 'json-schema',        '~> 2.8'
+gem 'manageiq-loggers',   '~> 0.1'
+gem 'manageiq-messaging', '~> 0.1.2'
+gem 'manageiq-password',  '~> 0.1'
+gem 'pg',                 '~> 1.0', :require => false
+gem 'prometheus-client',  '~> 0.8.0'
```
bdunne: Is this used?

eclarizio: I was actually going to ask, I don't think so? It got pulled in during a merge conflict. I think it can be removed.

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -8,9 +10,11 @@ def order
         service_plan = model.find(service_plan_id)
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
-        #TODO: Publish a message with service plan ordering messaging client and simply
-        # return the task id, let the messaging client handle updating the task.
-        order_service_plan_and_update_task(task, service_plan)
+        messaging_client.publish_message(
+          :service => "topological_inventory-orderer",
```
eclarizio: Woops, this will need to change to the "operations-openshift" naming

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -1,3 +1,5 @@
+require "manageiq-messaging"
```
bdunne: Should this require move down into the `messaging_client` method where it's used?

Gemfile
```diff
@@ -1,13 +1,14 @@
...
-gem 'manageiq-loggers',  '~> 0.1'
-gem 'manageiq-password', '~> 0.1'
-gem 'pg',                '~> 1.0', :require => false
-gem 'puma',              '~> 3.0'
-gem 'rack-cors',         '>= 0.4.1'
-gem 'rails',             '~> 5.2.2'
+gem 'jbuilder',           '~> 2.0'
+gem 'json-schema',        '~> 2.8'
+gem 'manageiq-loggers',   '~> 0.1'
+gem 'manageiq-messaging', '~> 0.1.2'
```
bdunne: Should this be `:require => false`?

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -40,6 +38,19 @@ def list_params
         params.permit(:source_id, :tenant_id, :service_offering_id)
       end
 
+      def messaging_client
+        ManageIQ::Messaging::Client.open(messaging_opts)
+      end
+
+      def messaging_opts
+        {
```
bdunne: Should this me memoized on the class or something?

spec/controllers/api/v0x0/service_plans_controller_spec.rb
```diff
@@ -25,35 +25,42 @@
...
 
-    before do
-      allow_any_instance_of(ServicePlan).to receive(:order).with(order_parameters).and_return("task_context")
-    end
-
     context "with a well formed service plan id" do
-      before do
-        payload = {
+      let(:client) { double(:client) }
+      let(:payload) do
```
bdunne: Can this and `order_parameters` above be deduplicated?

spec/controllers/api/v0x0/tasks_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+RSpec.describe Api::V0x0::TasksController, :type => :request do
+  it("Uses IndexMixin")   { expect(described_class.instance_method(:index).owner).to eq(Api::V0::Mixins::IndexMixin) }
+  it("Uses ShowMixin")    { expect(described_class.instance_method(:show).owner).to eq(Api::V0::Mixins::ShowMixin) }
+  it("Uses UpdateMixin")  { expect(described_class.instance_method(:update).owner).to eq(Api::V0::Mixins::UpdateMixin) }
+
+  let(:tenant) { Tenant.create! }
+
+  it "patch /tasks/:id updates a Task" do
```
bdunne: Is this test necessary?  It should be covered by the UpdateMixin spec

Gemfile
eclarizio: Yeah that's probably fine

app/controllers/api/v0/service_plans_controller.rb
eclarizio: I think that's fine, the specs fail after moving but requiring it in the specs makes them pass again.

spec/controllers/api/v0x0/tasks_controller_spec.rb
eclarizio: Doesn't that spec just test that it includes the `#update` method? This spec is making sure that if you pass in the status and context those get updated. I'd argue we could also additionally pass in another parameter that we expect *not* to get updated, like name or something.

eclarizio:
@bdunne Since you've been reviewing and have some good suggestions, do you know what I did wrong with my swagger doc that travis is complaining about? It seems to think the `PATCH` action is missing but I added it to both versions of the swagger doc. 😕

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,40 @@ def order
...
+      def messaging_opts
+        @messaging_opts ||= {
+          :protocol => :Kafka,
+          :host     => ENV["QUEUE_HOST"] || "localhost",
+          :port     => ENV["QUEUE_PORT"] || "9092",
+          :encoding => "json"
+        }
+      end
+
+      def model
```
carbonin: Why is this being added here? Wasn't there a call to `.model` before this PR?

app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,21 @@ module V0
...
+
+      def params_for_list
+        params.permit(:tenant_id)
+      end
+
+      def params_for_update
+        params.permit(:id, :status, :context)
+      end
+
+      def model
```
carbonin: I don't see this being used anywhere.

app/controllers/api/v0/service_plans_controller.rb
carbonin: Yeah, this is all defined here https://github.com/ManageIQ/topological_inventory-api/blob/master/app/controllers/application_controller.rb#L21-L23

It works as is:
```
irb(main):005:0> Api::V0::ServicePlansController.send(:model)
=> ServicePlan (call 'ServicePlan.connection' to establish a connection)
```

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,40 @@ def order
...
+      end
+
+      def messaging_client
+        require "manageiq-messaging"
+
+        ManageIQ::Messaging::Client.open(messaging_opts)
+      end
+
+      def messaging_opts
+        @messaging_opts ||= {
```
carbonin: Does this need to specify a topic?

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -5,12 +5,14 @@ class ServicePlansController < ApplicationController
...
       def order
-        service_plan = model.find(order_params[:service_plan_id].to_i)
+        service_plan = model.find(params_for_order[:service_plan_id].to_i)
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
-        #TODO: Publish a message with service plan ordering messaging client and simply
-        # return the task id, let the messaging client handle updating the task.
-        order_service_plan_and_update_task(task, service_plan)
+        messaging_client.publish_message(
+          :service => "platform.topological_inventory.operations-openshift",
```
carbonin: This should be `platform.topological-inventory.operations-openshift`

eclarizio:
@miq-bot remove_label wip

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,36 @@ def order
...
-      def order_params
+      def params_for_order
         params.permit(
           :service_plan_id,
           :service_parameters          => {},
           :provider_control_parameters => {}
         ).to_h
       end
+
+      def params_for_list
```
bdunne: Why is this being added back?

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,36 @@ def order
 
       private
 
-      def order_service_plan_and_update_task(task, service_plan)
-        task.context = service_plan.order(order_params.slice(:service_parameters, :provider_control_parameters))
-        task.status = "completed"
-        task.save!
+      def service_plan_id
+        params[:service_plan_id].to_i
```
bdunne: This needs to reuse `params_for_order` below.

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,36 @@ def order
...
       end
 
-      def order_params
+      def params_for_order
```
bdunne: Should this cache the hash to prevent generating multiple hashes (and warnings)

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -19,19 +21,36 @@ def order
...
           :service_parameters          => {},
           :provider_control_parameters => {}
         ).to_h
       end
+
+      def params_for_list
+        params.permit(:source_id, :tenant_id, :service_offering_id)
+      end
+
+      def messaging_client
```
bdunne: I know that this is only called once right now, but should this be cached to prevent opening multiple connections?

app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,17 @@ module V0
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
+
+      private
+
+      def params_for_list
```
bdunne: Same here

app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,17 @@ module V0
...
+        params.permit(:tenant_id)
+      end
+
+      def params_for_update
```
bdunne: And here

spec/controllers/api/v0x0/tasks_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+RSpec.describe Api::V0x0::TasksController, :type => :request do
```
bdunne: Can you add the same tests for v0.1?  This feels like it should be an independent PR.

app/controllers/api/v0/service_plans_controller.rb
eclarizio: I realized I wasn't using this anyway, so I just got rid of it

app/controllers/api/v0/service_plans_controller.rb
eclarizio: Sure, that's simple enough.

spec/controllers/api/v0x0/tasks_controller_spec.rb
eclarizio: Moved to #137

app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -5,12 +5,14 @@ class ServicePlansController < ApplicationController
...
-        service_plan = model.find(order_params[:service_plan_id].to_i)
+        service_plan = model.find(params_for_order[:service_plan_id].to_i)
         task = Task.create!(:tenant => service_plan.tenant, :status => "started")
 
-        #TODO: Publish a message with service plan ordering messaging client and simply
-        # return the task id, let the messaging client handle updating the task.
-        order_service_plan_and_update_task(task, service_plan)
+        messaging_client.publish_message(
+          :service => "platform.topological-inventory.operations-openshift",
+          :message => "order_service",
```
agrare: Can this be a little more structured?  Like `ServicePlan.order` to make it easier to process generally on the operations side?

PR #104, opened 2019-01-25
Enforce tenancy

gtanzillo:
First pass at enforcing tenancy

Tenancy enforcement requires the environment variable `ENFORCE_TENANCY` to be set to "true". Otherwise no enforcement will be done. The default behavior is NOT enforce tenancy.

https://projects.engineering.redhat.com/browse/TPINVTRY-127

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,48 @@ class ApplicationController < ActionController::API
   end
 
   private
+  set_current_tenant_through_filter
+  before_action :set_the_current_tenant
+
+  def set_the_current_tenant
+    return unless ENV["ENFORCE_TENANCY"]
+
+    tenant = Tenant.find_by(:external_tenant => user_identity)
+    if tenant
+      set_current_tenant(tenant)
+    else
+      error_document = {
+          "errors" => [
+              {
+                  "status" => "401",
+                  "detail" => "Account number not present or not found"
+              }
+          ]
+      }
+      render :json => error_document, :status => :unauthorized
+    end
+  end
+
+  def user_account_number
+    Base64.decode64(request.headers.fetch_path("x-rh-identity", "identity", "account_number"))
+  end
+
+  def user_identity
+    # x-rh-identity = {
+    #     "identity" => {
+    #         "account_number" => 123456,
+    #         "type" => "String"
+    #     },
+    #     "user" => {},
+    #     "system" => {},
+    #     "internal" => {},
+    # }
+    ident_key = "x-rh-identity"
+    return unless request.headers.key?(ident_key)
+
+    ident = JSON.parse(Base64.decode64(request.headers[ident_key]))
+    ident.fetch_path("identity", "account_number")
+  end
 
   private_class_method def self.model
     @model ||= controller_name.classify.constantize
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::API
   before_action :set_the_current_tenant
 
   def set_the_current_tenant
-    return unless ENV["ENFORCE_TENANCY"]
+    return unless Tenant.tenancy_enabled?
 
     tenant = Tenant.find_by(:external_tenant => user_identity)
     if tenant
@@ -29,19 +29,15 @@ class ApplicationController < ActionController::API
     end
   end
 
-  def user_account_number
-    Base64.decode64(request.headers.fetch_path("x-rh-identity", "identity", "account_number"))
-  end
-
   def user_identity
-    # x-rh-identity = {
-    #     "identity" => {
-    #         "account_number" => 123456,
-    #         "type" => "String"
-    #     },
+    # "x-rh-identity" => {
+    #   "identity" => {
+    #     "account_number" => 123456,
     #     "user" => {},
     #     "system" => {},
     #     "internal" => {},
+    #     "type" => "String"
+    #   }
     # }
     ident_key = "x-rh-identity"
     return unless request.headers.key?(ident_key)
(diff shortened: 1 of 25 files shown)


agrare:
cc @bsorota

spec/controllers/api/v0x0/application_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  let(:external_tenant) { rand(1000).to_s }
+  let(:identity)        { Base64.encode64({'identity' => { 'account_number' => external_tenant}}.to_json) }
+
+  context "with tenancy enforcement" do
+    after { controller.send(:set_current_tenant,  nil) }
+
+    it "get /source with tenant" do
+      headers = { "CONTENT_TYPE" => "application/json", "x-rh-auth-identity" => identity }
+
+      with_modified_env :ENFORCE_TENANCY => "true" do
```
Fryguy: Since you are wrapping the entire test in an ENV change, then you can just use the built-in rspec stub_const instead of introducing a gem.  For example,

```
context "with tenancy enforcement" do
  before { stub_const("ENFORCE_TENANCY", true) }
  ...
end
```

spec/controllers/api/v0x0/application_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
+RSpec.describe ApplicationController, :type => :request do
+  include SpecHelper
+
+  let(:tenant)          { Tenant.create!(:external_tenant => external_tenant) }
+  let(:source_type)     { SourceType.create!(:name => "openshift", :product_name => "OpenShift", :vendor => "Red Hat") }
+  let(:source)          { Source.create!(:source_type_id => source_type.id.to_s, :tenant_id => tenant.id , :name => "abc", :uid => "123") }
```
Fryguy: source_type.id should not have a .to_s on it.

spec/controllers/api/v0x0/application_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  let(:external_tenant) { rand(1000).to_s }
+  let(:identity)        { Base64.encode64({'identity' => { 'account_number' => external_tenant}}.to_json) }
```
Fryguy: @bdunne @syncrou When we get to the user stuff...this feels like a method that should live in ManageIQ::API::Common so we don't have to reinvent it everywhere.

spec/controllers/api/v0x0/application_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    it "get /sources" do
+      headers = { "CONTENT_TYPE" => "application/json" }
+
+      get("/api/v0.0/sources", :headers => headers)
+
+      expect(response.status).to eq(200)
+    end
+  end
+end
```
Fryguy: Trailing newline missing.

spec/controllers/api/v0x0/application_controller_spec.rb
syncrou: @Fryguy - Yes, I have a method called `ManageIQ::API::Common::Users.encode that does exactly this. Hoping to push up that PR later today.

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,40 @@
...
+    # x-rh-auth-identity = {
+    #     "identity" => {
+    #         "account_number" => 123456,
+    #         "type" => "String"
+    #     },
+    #     "user" => {},
+    #     "system" => {},
+    #     "internal" => {},
+    # }
+    ident_key = "x-rh-auth-identity"
```
syncrou: @gtanzillo In the approved IPP the key is now:  'x-rh-identity'

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,41 @@
...
+
+    tenant = Tenant.find_by(:external_tenant => ident)
+    if tenant
+      set_current_tenant(tenant)
+    else
+      render :json => { :errors => "Unauthorized" }, :status => :unauthorized
+    end
+  end
+
+  def user_account_number
```
syncrou: @gtanzillo - It appears this method is no longer in use.  It was also incorrectly calling decode on a nested key. Line 36 decodes correctly.

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,41 @@
...
+    Base64.decode64(request.headers.fetch_path("x-rh-identity", "identity", "account_number"))
+  end
+
+  def user_identity
```
bdunne: Should this be cached?

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,41 @@
...
+
+  def set_the_current_tenant
+    return unless ENV["ENFORCE_TENANCY"]
+    return unless ident = user_identity
+
+    tenant = Tenant.find_by(:external_tenant => ident)
+    if tenant
+      set_current_tenant(tenant)
+    else
+      render :json => { :errors => "Unauthorized" }, :status => :unauthorized
```
bdunne: The error document should probably be a little bit more descriptive.  See https://jsonapi.org/format/#errors and https://github.com/ManageIQ/topological_inventory-api/pull/109/files#diff-df731b83f826451eea12d418a5da8d51R8 for an example

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,41 @@
 class ApplicationController < ActionController::API
 
   private
+  set_current_tenant_through_filter
+  before_action :set_the_current_tenant
+
+  def set_the_current_tenant
+    return unless ENV["ENFORCE_TENANCY"]
+    return unless ident = user_identity
```
bdunne: This `return` would prevent us from getting the error response, right?

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,41 @@
...
+
+    tenant = Tenant.find_by(:external_tenant => ident)
```
bdunne: Should this `find_or_create_by`? as long as it's not `nil`?

app/controllers/application_controller.rb
gtanzillo: We're going to move this to a common gem so we can take care of that there

gtanzillo:
Closing in place of https://github.com/ManageIQ/topological_inventory-api/pull/165

PR #105, opened 2019-01-29
Allow for path-only redirect links in the router

bdunne:
This is necessary for the 3scale API gateway to allow redirects.  The location can't include the hostname.

### DIFF
--- a/config/initializers/action_dispatch-routing-option_redirect.rb
+++ b/config/initializers/action_dispatch-routing-option_redirect.rb
@@ -1,0 +1,20 @@
+module OptionRedirectEnhancements
+  def serve(req)
+    uri = URI.parse(path(req.path_parameters, req))
+
+    req.commit_flash
+
+    body = %(<html><body>You are being <a href="#{ERB::Util.unwrapped_html_escape(uri.to_s)}">redirected</a>.</body></html>)
+
+    headers = {
+      "Location" => uri.to_s,
+      "Content-Type" => "text/html",
+      "Content-Length" => body.length.to_s
+    }
+
+    [ status, headers, [body] ]
+  end
+end
+
+require 'action_dispatch/routing/redirection'
+ActionDispatch::Routing::OptionRedirect.prepend(OptionRedirectEnhancements)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -8,7 +8,7 @@ Rails.application.routes.draw do
   end
 
   scope :as => :api, :module => "api", :path => prefix do
-    match "/v0/*path", :via => [:delete, :get, :options, :patch, :post], :to => redirect("#{prefix}/v0.0/%{path}")
+    match "/v0/*path", :via => [:delete, :get, :options, :patch, :post], :to => redirect(:path => "/#{prefix}/v0.0/%{path}", :only_path => true)
 
     namespace :v0x1, :path => "v0.1" do
       resources :authentications,         :only => [:create, :destroy, :index, :show, :update]


bdunne:
Specifically, see the `Location`:
```
< HTTP/1.1 301 Moved Permanently
< Location: http://localhost:3000/api/v0.0/sources
```
vs
```
< HTTP/1.1 301 Moved Permanently
< Location: /api/v0.0/sources?limit=1
```

Before:
```
$ curl -Lvv localhost:3000/api/v0/sources?limit=1
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET /api/v0/sources?limit=1 HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Location: http://localhost:3000/api/v0.0/sources
< Content-Type: text/html
< Cache-Control: no-cache
< X-Request-Id: 1f326429-b74b-4a6d-83e2-21a6cb602072
< X-Runtime: 0.067160
< Vary: Origin
< Content-Length: 104
< 
* Ignoring the response-body
* Connection #0 to host localhost left intact
* Issue another request to this URL: 'http://localhost:3000/api/v0.0/sources'
* Found bundle for host localhost: 0x558715d0e790 [can pipeline]
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET /api/v0.0/sources HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< ETag: W/"939acee49a805283c1044f67276456e4"
< Cache-Control: max-age=0, private, must-revalidate
< X-Request-Id: 86279863-2ce4-42d9-b3f3-f600d19d17e7
< X-Runtime: 0.168669
< Vary: Origin
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
```

After:
```
$ curl -Lvv localhost:3000/api/v0/sources?limit=1
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET /api/v0/sources?limit=1 HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Location: /api/v0.0/sources?limit=1
< Content-Type: text/html
< Cache-Control: no-cache
< X-Request-Id: 0ab45d03-9d0b-4a0a-882a-556ddcb9d740
< X-Runtime: 0.072721
< Vary: Origin
< Content-Length: 91
< 
* Ignoring the response-body
* Connection #0 to host localhost left intact
* Issue another request to this URL: 'http://localhost:3000/api/v0.0/sources?limit=1'
* Found bundle for host localhost: 0x55a2fc4f8790 [can pipeline]
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET /api/v0.0/sources?limit=1 HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< ETag: W/"939acee49a805283c1044f67276456e4"
< Cache-Control: max-age=0, private, must-revalidate
< X-Request-Id: 71c8de9e-07fc-4163-af98-17b6025a2af2
< X-Runtime: 0.208527
< Vary: Origin
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
```

bdunne:
@abellotti Please review

abellotti:
in routes.rb,  is prefix based on the listening path, i.e. would it be /r/insights/platform/topological-inventory/  when this puppy is answering external api calls ?

bdunne:
> would it be /r/insights/platform/topological-inventory/ when this puppy is answering external api calls?

@abellotti yes, it would

abellotti:
LGTM!! 👍

carbonin:
I'd really like a better way to do this. It doesn't look like the `only_path` option is really supported in redirect so really we're just hacking in our own behavior.
ref:
https://api.rubyonrails.org/classes/ActionDispatch/Routing/Redirection.html#method-i-redirect

@abellotti can you give a bit more background on why we need to omit the host in the redirect header? From what I'm reading this is usually handled at the web-server layer (which I guess in this case is 3scale as it's proxying all the requests).

abellotti:
host that you see/get is not the real host, that information is not forwarded to us (gateway/3scale) by the edge servers and proxies, an api call can be api.access.ci..., could be cert-api.access.ci..., etc.  putting the host ourselves in the LOCATION would cause the redirect to fail since it's really incorrect.  so when eliminated (just having the /r/insights/... path), the clients (curl/libs/etc) rebuilds it with the original host/port/creds (that the micro-service won't have anyways)/etc. which works like a champ.

PR #108, opened 2019-01-30
[WIP] Allow to filter by archived at

Ladas:
So adding this will allow us to do:

```
# Show only active entities
curl  http://localhost:3000/api/v0.1/service_offerings?archived_at= | python -mjson.tool
# Show all entities 
curl  http://localhost:3000/api/v0.1/service_offerings | python -mjson.tool
```


Ladas:
replaced by https://github.com/ManageIQ/topological_inventory-api/pull/111

PR #113, opened 2019-02-05
Add /internal/v0.0/authentications/:id api

bdunne:
- Add redirect for /internal/v0/* to /internal/v0.0/*
- Upgrade to manageiq-password v0.2.1 with corrected .encrypted_columns
- Allow `GET /internal/v0.0/authentications/1?expose_encrypted_attribute[]=password` to merge the unencrypted password into the response object
- For security purposes it's only on 1 instance and you must know the name of the attribute you're asking for and it's not exposed through the 3scale gateway, so you must hit the service directly.

https://projects.engineering.redhat.com/browse/TPINVTRY-165

### DIFF
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -5,17 +5,20 @@ module OpenApi
       encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
       encryption_filtered = previous.except(*encrypted_columns_set)
       return encryption_filtered unless arg.key?(:prefixes)
-      /\A\/?(?<api>\w+)\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ arg[:prefixes].first
-      version = [major, minor].compact.join(".")
+      version = api_version_from_prefix(arg[:prefixes].first)
       schema  = Api::Docs[version].definitions[self.class.name]
       attrs   = encryption_filtered.slice(*schema["properties"].keys)
-
       schema["properties"].keys.each do |name|
         attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
         attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
         attrs[name] = self.public_send(name) if !attrs.key?(name) && !encrypted_columns_set.include?(name)
       end
       attrs.compact
+    end
+
+    def api_version_from_prefix(prefix)
+      /\/?\w+\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ prefix
+      [major, minor].compact.join(".")
     end
   end
 end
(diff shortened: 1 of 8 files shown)


lib/open_api/serializer.rb
```diff
@@ -5,7 +5,7 @@ def as_json(arg = {})
       encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
       encryption_filtered = previous.except(*encrypted_columns_set)
       return encryption_filtered unless arg.key?(:prefixes)
-      /\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ arg[:prefixes].first
+      /\A\/?(?<api>\w+)\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ arg[:prefixes].first
```
gtanzillo: Not sure I understand what this change is doing. Seems like the old expression would match both `/internal/v0.0/authentications/1` and `/api/v0.0/authentications/1`

bdunne: The old regex did not allow for a leading `/`.  So, `/api/v0.0/authentications/1` would not be parsed properly, but `api/v0.0/authentications/1` would.

gtanzillo: Got it, thanks 👍

bdunne:
> Are you planning on adding tests for the new controller?

Tests added

lib/open_api/serializer.rb
carbonin: Will this break if we add additional path components prior to the version? For example `/r/insights/platform/topological-inventory/v0.0/...`

I feel like the multiple path components won't match the new `\w+` for the `api` capture group.

spec/swagger_spec.rb
```diff
@@ -29,7 +29,8 @@
 
       it "matches the routes" do
         redirect_routes = [{:path => "#{path_prefix}/#{app_name}/v0/*path", :verb => "DELETE|GET|OPTIONS|PATCH|POST"}]
-        expect(rails_routes).to match_array(swagger_routes + redirect_routes)
+        internal_api_routes = [{:path => "/internal/v0/*path", :verb => "GET"}, {:path=>"/internal/v0.0/authentications/:id", :verb=>"GET"}]
+        expect(rails_routes).to match_array(swagger_routes + redirect_routes + internal_api_routes)
```
carbonin: Does this mean that we are not creating a swagger spec with the internal routes? Is the intention for us not to be able to generate a client for those? Or maybe we should create a separate spec?

lib/open_api/serializer.rb
bdunne: Fixed and added tests.

bdunne:
> I'm guessing the limitation to 1 instance and not being exposed through the 3scale gateway is a configuration setting somewhere else?

3scale will only send us requests starting with `/r/insights...` and this is listening to `/internal`.

spec/lib/open_api/serializing_spec.rb
```diff
@@ -25,5 +25,14 @@ def to_hash
 
       expect(model.new.as_json).to eq("a" => 1, "b" => 2, "secret" => "value")
     end
+
+    it "works" do
```
carbonin: Ehhh can we put something a bit better here?

Ideally this would read `#api_version_from_prefix parses the correct version`

PR #114, opened 2019-02-06
Add /v*/openapi.json to fetch the swagger yaml

bdunne:
The IPP now requires that the api document be available at `/api/v0.1/openapi.json`

### DIFF
--- a/public/doc/swagger-2-v0.1.0.yaml
+++ b/public/doc/swagger-2-v0.1.0.yaml
@@ -469,6 +469,15 @@ paths:
             "$ref": "#/definitions/Flavor"
         404:
           description: Not found
+  "/openapi.json":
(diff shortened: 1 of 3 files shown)


public/doc/swagger-2-v0.1.0.yaml
```diff
@@ -469,6 +469,15 @@ paths:
             "$ref": "#/definitions/Flavor"
         404:
           description: Not found
+  "/openapi.json":
```
Fryguy: Should this be in the swagger doc?

bdunne: ¯\\\_(ツ)_/¯ It's a valid path, and the tests will fail unless I add this to an exception list

PR #115, opened 2019-02-07
Fix allowed create/update params for endpoints

agrare:
Endpoints were missing verify_ssl and certificate_authority as allowed
create/update params.


agrare:
cc @bdunne @martinpovolny

PR #118, opened 2019-02-07
[WIP] Add a request tracking middleware

bdunne:
Example output:
```
[----] I, [2019-02-07T17:35:39.416597 #21395:2b17dfb91078]  INFO -- : {:request_uri=>"/api/v0.0/sources/1", :request_method=>"GET", :request_duration=>0.190625771, :response_code=>200, :request_source=>"127.0.0.1"}
```

### DIFF
--- a/app/middleware/tracker_middleware.rb
+++ b/app/middleware/tracker_middleware.rb
@@ -1,0 +1,27 @@
+class TrackerMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    dup._call(env)
+  end
+
+  def _call(env)
+    request_started_on = Time.now
+    status, headers, response = @app.call(env)
+    request_finished_on = Time.now
+
+    data = {
+      :request_duration => (request_finished_on - request_started_on),
+      :request_method   => env["REQUEST_METHOD"], # GET, POST, PATCH, etc.
+      :request_source   => env["REMOTE_ADDR"],
+      :request_uri      => env["REQUEST_URI"], # Path + Query
+      :response_code    => status,
+    }
+
+    Rails.logger.info(data.inspect)
+
+    [status, headers, response]
+  end
+end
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -44,5 +44,8 @@ class Application < Rails::Application
 
     ManageIQ::API::Common::Logging.activate(config)
     ManageIQ::API::Common::Metrics.activate(config, "topological_inventory_api")
+
+    require Rails.root.join("app/middleware/tracker_middleware").to_s
```
Fryguy: Why app/middleware, as opposed to lib/middleware?

Fryguy: Also we shouldn't be using require on things in autoload dirs.  At a minimum we should use `require_dependency`, or just let autoload load it (if possible)

app/middleware/tracker_middleware.rb
```diff
@@ -0,0 +1,27 @@
+class TrackerMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    dup._call(env)
```
Fryguy: Why the .`dup`?

app/middleware/tracker_middleware.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def _call(env)
+    request_started_on = Time.now
```
Fryguy: This should probably be in UTC for consistency.

config/application.rb
```diff
@@ -44,5 +44,8 @@ class Application < Rails::Application
...
+    config.middleware.use(::TrackerMiddleware)
```
Fryguy: For tracking you probably want to make sure this is in a consistent location in the stack (probably the top of the stack.  So, you want want to use .insert instead of .use

app/middleware/tracker_middleware.rb
bdunne: I thought that was a good idea after reading https://ieftimov.com/writing-rails-middleware see `Thread Safety`

PR #126, opened 2019-02-12
Allow for subcollection requests on a polymorphic relation

bdunne:
- primary_collection_id and primary_collection_type need to be added to the safe_params
- :query was a mistake and should be removed

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,8 +42,19 @@ class ApplicationController < ActionController::API
   end
 
   def safe_params_for_list
-    # :limit & :query can be passed in for pagination purposes, but shouldn't show up as params for filtering purposes
-    params.permit(*api_doc_definition.all_attributes + [:limit, :query, :offset])
+    # :limit & :offset can be passed in for pagination purposes, but shouldn't show up as params for filtering purposes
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -62,6 +62,18 @@ def params_for_update
...
+  def request_path
+    request.env["REQUEST_URI"]
+  end
+
+  def request_path_parts
+    request_path.match(/\/(?<full_version_string>v\d+.\d+)\/(?<primary_collection_name>\w+)\/?(?<primary_collection_id>\d+)?\/?(?<subcollection_name>\w+)?/) || {}
+  end
+
+  def allowed_params_for_primary_collection
+    ["#{request_path_parts[:primary_collection_name].singularize}_id", "#{request_path_parts[:primary_collection_name].singularize}_type"]
```
agrare: @bdunne when I test this with `sources/availabilities` (https://github.com/ManageIQ/topological_inventory-api/pull/127/commits/49f52fd02a0aca4b248762b5ec5baa88ec44848d) this results in `["source_id", "source_type"]` when it would have to be `["resource_id", "resource_type"]` with `resource_type = Source` to work.

PR #127, opened 2019-02-12
Add Availabilities

agrare:
Depends on:

- [x] https://github.com/ManageIQ/topological_inventory-core/pull/98/files

### DIFF
--- a/public/doc/swagger-2-v0.0.1.yaml
+++ b/public/doc/swagger-2-v0.0.1.yaml
@@ -837,6 +837,20 @@ paths:
           description: Source deleted
         404:
           description: Not found
+  "/sources/{id}/availabilities":
+    get:
+      summary: List Availabilities for Source
+      operationId: listSourceAvailabilities
+      description: Returns an array of Availability objects
+      produces:
+      - application/json
+      responses:
+        200:
+          description: Availabilities array
+          schema:
+            type: array
+            items:
+                "$ref": "#/definitions/Availability"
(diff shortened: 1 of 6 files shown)


public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -837,6 +837,20 @@ paths:
           description: Source deleted
         404:
           description: Not found
+  "/sources/{id}/availabilities":
```
bdunne: Do we need to add this to the v0.0 api?

agrare: Probably not

agrare: Removed

public/doc/swagger-2-v0.1.0.yaml
```diff
@@ -2200,9 +2276,6 @@ definitions:
         readOnly: true
         title: Name
         type: string
-      resource_version:
```
bdunne: This shouldn't be removed

agrare: Hm I assumed this was missing from the core models but it is there on ServicePlans, need to look at why the generator cleared this.

agrare: Fixed

bdunne: Maybe you didn't have the latest version of `topological_inventory-core` when the generator was run?

PR #129, opened 2019-02-13
Generate definitions based on models

Ladas:
A dev tool for making it easier to keep 'provider models' in sync with API. This tool should only help us with the API yaml generation, it doesn't mean there has to be 1:1 API to models mapping in all cases.

### DIFF
--- a/lib/tasks/swagger_generate.rake
+++ b/lib/tasks/swagger_generate.rake
@@ -133,7 +133,9 @@ class SwaggerGenerator
   def swagger_definition_properties(klass_name)
     model = klass_name.constantize
     model.columns_hash.map do |key, value|
-      next if GENERATOR_BLACKLIST_ATTRIBUTES.include?(key.to_sym)
+      unless(GENERATOR_ALLOW_BLACKLISTED_ATTRIBUTES[key.to_sym] || []).include?(klass_name)
+        next if GENERATOR_BLACKLIST_ATTRIBUTES.include?(key.to_sym)
+      end
 
       [key, swagger_definition_properties_value(klass_name, model, key, value)]
     end.compact.sort.to_h
@@ -356,16 +358,17 @@ class SwaggerGenerator
   end
 end
 
-
-GENERATOR_BLACKLIST_ATTRIBUTES            = [
+GENERATOR_BLACKLIST_ATTRIBUTES           = [
   :resource_timestamp, :resource_timestamps, :resource_timestamps_max, :tenant_id
 ].to_set.freeze
+GENERATOR_ALLOW_BLACKLISTED_ATTRIBUTES   = {
+  :tenant_id => ['Source', 'Endpoint', 'Authentication'].to_set.freeze
+}
 GENERATOR_WHITELIST_PROVIDER_DEFINITIONS = [
   'Container', 'ContainerGroup', 'ContainerImage', 'ContainerNode', 'ContainerProject', 'ContainerTemplate', 'Flavor',
   'OrchestrationStack', 'ServiceInstance', 'ServiceOffering', 'ServiceOfferingIcon', 'ServicePlan', 'Tag', 'Tagging',
   'Vm', 'Volume', 'VolumeAttachment', 'VolumeType'
 ].to_set.freeze
-
 
 namespace :swagger do
   desc "Generate the swagger.yml contents"
--- a/lib/tasks/swagger_generate.rake
+++ b/lib/tasks/swagger_generate.rake
@@ -136,13 +136,26 @@ class SwaggerGenerator
 
   def swagger_definition_properties(klass_name)
     model = klass_name.constantize
-    model.columns_hash.map do |key, value|
+    attrs = model.columns_hash.map do |key, value|
       unless(GENERATOR_ALLOW_BLACKLISTED_ATTRIBUTES[key.to_sym] || []).include?(klass_name)
         next if GENERATOR_BLACKLIST_ATTRIBUTES.include?(key.to_sym)
       end
 
       [key, swagger_definition_properties_value(klass_name, model, key, value)]
-    end.compact.sort.to_h
+    end
+
+    if model.respond_to?(:taggable?) && model.taggable?
+      attrs << [
+        "taggings",
+        {
+          "type"     => "array",
+          "items"    => {"$ref" => "#/definitions/Tagging"},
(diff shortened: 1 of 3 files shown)


lib/tasks/swagger_generate.rake
```diff
@@ -123,6 +123,67 @@ class SwaggerGenerator
...
 
+  def swagger_definition(klass_name)
+    {
+      "type"       => "object",
+      "properties" => swagger_definition_properties(klass_name),
+    }
+  end
+
+  def swagger_definition_properties(klass_name)
+    model = klass_name.constantize
```
bdunne: Does this need to `safe_constantize` instead?  There may be things that are not backed by models.

Ladas: I'd probably let it explode. And when it does, we'll need to figure out what to do with these. We'll probably let them to be defined manually?

bdunne: That's fine for now.

bdunne:
Only issue I see is that the Tagging definition is not used... image

Although...
```
$ swagger-cli validate public/doc/swagger-2-v0.1.0.yaml 
public/doc/swagger-2-v0.1.0.yaml is valid
```

Fryguy:
@bdunne Isn't this exactly what the original code had and we backed away because they had things that were not backed by models? cc @syncrou

bdunne:
No, @Fryguy I think you're thinking of [this](https://github.com/ManageIQ/topological_inventory-api/blob/master/spec/swagger_spec.rb#L120)

syncrou:
@bdunne, @Fryguy yes, that was mentioned [here](https://github.com/ManageIQ/service_portal-api/pull/55/files#r241899649). Service Portal does have non model backed services in the OpenAPI spec e.g. https://github.com/ManageIQ/service_portal-api/blob/master/public/service_portal/v0.1.0/openapi.json#L1194-L1162

Fryguy:
@bdunne I was thinking more of what @syncrou shared...that is, having a 1-1 mapping between swagger endpoint and model I think is a bad idea, because we may choose to create endpoints that are an aggregation or an abstraction that aren't backed by literal models.  To me, an API is a presentation layer, and you typically don't map your backend models directly to presentation.

Ladas:
@Fryguy we should think of this only as a tool for helping us to maintain the yaml. There are already sections added manually. But for those that share a model, it's nice that we don't have to write them manually (as you can see we were missing a lot of attributes, sorting was random, etc.)

lib/tasks/swagger_generate.rake
```diff
@@ -295,6 +358,18 @@ class SwaggerGenerator
   end
 end
 
+GENERATOR_BLACKLIST_ATTRIBUTES           = [
+  :resource_timestamp, :resource_timestamps, :resource_timestamps_max, :tenant_id
+].to_set.freeze
+GENERATOR_ALLOW_BLACKLISTED_ATTRIBUTES   = {
+  :tenant_id => ['Source', 'Endpoint', 'Authentication'].to_set.freeze
+}
+GENERATOR_WHITELIST_PROVIDER_DEFINITIONS = [
```
agrare: Wonder if we could find a better name for this?  `GENERATOR_WHITELIST_PROVIDER_DEFINITIONS` sounds to me like we allow these things and default drop everything else.

It looks like it really is just a list of type that cannot be created through this API (aka populated by refresh).  How about `GENERATOR_READ_ONLY_DEFINITIONS` ?

Ladas:
@agrare @bdunne so I've added the `taggings` definition to all `.taggable?`, so now this should be the same yaml as before, except all the missing attributes were added and sorted. So does this look ready to you?

spec/requests/api/v0.1/source_types_spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe("v0.1 - SourceTypes") do
-  let(:attributes)      { {"name" => "test_name", "product_name" => "Test Product", "vendor" => "TestVendor"} }
+  let(:attributes)      { {"name" => "test_name", "product_name" => "Test Product", "vendor" => "TestVendor", "created_at"=>"2019-02-13T16:23:24Z", "updated_at"=>"2019-02-13T16:23:24Z"} }
```
bdunne: I don't think these should be added to the attributes.  This is used for record creation and I don't think we should ever pass values for these columns.

Ladas: ok, so I'll need to tweak the asserts

Ladas: hm right, this also means that these columns should always be readOnly :-)

Ladas: @bdunne hm, I've added the readOnly and the spec is still passing. The spec for `post` should fail right? Since we should filter out the `readOnly` attrs?

bdunne: We don't fail for read-only attributes on create.  There's no concept in swagger for things allowed for create, but read-only after that.  We would have to duplicate every object definition that allows for create and didn't want the overhead right now.

Ladas: hm ok, that make sense, we might need to have different concept for non writable cols (like a separate definition for writable, then readable will inherit that and add more?)

bdunne: Don't forget about PATCH / update

mkanoor:
@Ladas @Fryguy 
This PR broke the create_json_schema attribute in the ServicePlan object. It should have a datatype of object instead of string.

I have a manual fix for this in PR #160 
Can this tool be fixed to sort this issue, otherwise it will keep breaking it every time the code is used to generate the yams file.

Fryguy:
@mkanoor Can you open a Github issue so we don't lose track of this

Ladas:
@Fryguy @mkanoor responded in https://github.com/ManageIQ/topological_inventory-api/pull/160, I've fixed the generator while ago.

mkanoor:
@Ladas 
Can you point to a PR when that was fixed. I think @abellotti generated the yaml recently. @abellotti can you confirm if we are using the latest generator code

Ladas:
@mkanoor https://github.com/ManageIQ/topological_inventory-api/pull/141 , but as I mentioned, the generator doesn't change the jsonb type, once it is set

PR #131, opened 2019-02-14
Replace IDReadOnly with ID and a readOnly property

eclarizio:
I ran into an issue on https://github.com/ManageIQ/topological_inventory-operations-openshift/pull/13 with specs where the model definition returned from a `topological_inventory-api-client` call was returning objects without the ID components. It turns out that it was trying to return a sub-model called `IDReadOnly` instead of a simple string, so I wasn't able to actually get the ID of anything.

I think something is wonky in how the openapi-generator ends up interpreting the swagger yaml, as I found a few issues with the way the `readOnly` property works documented [here](https://github.com/swagger-api/swagger-editor/issues/529) (most importantly, [this comment that explains we can't use both `readOnly` and `$ref`](https://github.com/swagger-api/swagger-editor/issues/529#issuecomment-288271664))

There's also [this issue](https://github.com/swagger-api/swagger-codegen/issues/6537) that was opened and remains open. There is [a comment about fixing references that use readOnly](https://github.com/swagger-api/swagger-codegen/issues/6537#issuecomment-341490912) and other places that don't seem to be using it correctly in a PR, which has been merged, but the latest comment seems to insinuate that the problem is still happening.

~~So, for now, I'm simply splitting up readonly ids (which is the default, `ID`), and non-readonly ids (now `UpdateableID`, if there's another name we should go with then let me know).~~ I'm still not entirely sure that `readOnly` even does anything, at least from a codegen perspective, because when I run the openapi-generator after just having changed the `readOnly` property, there are no changes.

@bdunne Do you have any insight into this?

### DIFF
--- a/public/doc/swagger-2-v0.1.0.yaml
+++ b/public/doc/swagger-2-v0.1.0.yaml
@@ -1341,9 +1341,9 @@ definitions:
     type: object
     properties:
       id:
-        "$ref": "#/definitions/IDReadOnly"
+        "$ref": "#/definitions/ID"
       tenant_id:
-        "$ref": "#/definitions/IDReadOnly"
+        "$ref": "#/definitions/ID"
       authtype:
         type: string
         example: openshift_default
@@ -1396,9 +1396,9 @@ definitions:
     type: object
     properties:
       id:
-        "$ref": "#/definitions/IDReadOnly"
+        "$ref": "#/definitions/ID"
       tenant_id:
-        "$ref": "#/definitions/IDReadOnly"
+        "$ref": "#/definitions/ID"
       name:
         type: string
         readOnly: true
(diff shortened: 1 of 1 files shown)


public/doc/swagger-2-v0.1.0.yaml
```diff
@@ -1354,7 +1354,7 @@ definitions:
         type: string
         example: Endpoint
       resource_id:
-        "$ref": "#/definitions/ID"
+        "$ref": "#/definitions/UpdateableID"
```
bdunne: I think this can change to `ID` as well, then we don't need `UpdateableID`

eclarizio: Ah ok, sounds good, it was the only one that wasn't originally set to `IDReadOnly` so I kept the consistency. We can always re-add `UpdateableID` if this or some other relational object id should be updateable.

PR #137, opened 2019-02-25
Add task updating patch endpoint

eclarizio:
As suggested by @bdunne in #103, I've separated out the task updating into a separate PR. #103 will need to be rebased after this goes in.

### DIFF
--- a/app/controllers/api/v0/tasks_controller.rb
+++ b/app/controllers/api/v0/tasks_controller.rb
@@ -4,20 +4,6 @@ module Api
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
       include Api::V0::Mixins::UpdateMixin
-
-      private
-
-      def list_params
-        params.permit(:tenant_id)
-      end
-
-      def update_params
-        params.permit(:status, :context)
-      end
-
-      def model
-        Task
-      end
     end
   end
 end
--- a/app/controllers/api/v0/tasks_controller.rb
+++ b/app/controllers/api/v0/tasks_controller.rb
@@ -3,6 +3,21 @@ module Api
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
+
+      private
+
+      def list_params
+        params.permit(:tenant_id)
+      end
+
+      def update_params
+        params.permit(:status, :context)
+      end
+
+      def model
+        Task
+      end
     end
   end
 end
--- a/app/controllers/api/v0/tasks_controller.rb
+++ b/app/controllers/api/v0/tasks_controller.rb
@@ -3,7 +3,6 @@ module Api
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
-      include Api::V0::Mixins::UpdateMixin
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,17 @@ module V0
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
+
+      private
+
+      def params_for_list
```
bdunne: These methods shouldn't be needed, the generic ones in `ApplicationController` should handle it

spec/controllers/api/v0x0/tasks_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  it("Uses IndexMixin")   { expect(described_class.instance_method(:index).owner).to eq(Api::V0::Mixins::IndexMixin) }
+  it("Uses ShowMixin")    { expect(described_class.instance_method(:show).owner).to eq(Api::V0::Mixins::ShowMixin) }
+  it("Uses UpdateMixin")  { expect(described_class.instance_method(:update).owner).to eq(Api::V0::Mixins::UpdateMixin) }
+
+  let(:tenant) { Tenant.create! }
+
+  it "patch /tasks/:id updates a Task" do
+    task = Task.create!(:status => "started", :context => "context1", :tenant => tenant)
+
+    patch(api_v0x0_task_url(task.id), :params => {:status => "completed", :context => "context2"})
```
bdunne: I think the params need a `to_json` in order to be properly parsed from the body

eclarizio: Ah good call, that was why I needed the `params_for_update` before, without the `to_json` the spec passed with `params_for_update`. Will fix.

config/routes.rb
```diff
@@ -119,7 +119,7 @@ def put(*_args); end
         resources :volumes,                 :only => [:index]
       end
       resources :tags, :only => [:index, :show]
-      resources :tasks, :only => [:index, :show]
+      resources :tasks, :only => [:index, :show, :update]
```
bdunne: This should be in the v0.1 API, not v0.0, right?

public/doc/swagger-2-v0.0.1.yaml
```diff
@@ -1121,6 +1121,27 @@ paths:
             "$ref": "#/definitions/Task"
         404:
           description: Not found
+    patch:
```
bdunne: I don't think we should be updating this version of the API any more.

eclarizio: Good idea, on the contrary, should we be bumping the version by a patch level or something then for this change? So instead of modifying v0.1.0, should I be making a new v0.1.1 with the patch change?

bdunne: Not now, I think we need to figure out the deployment strategy first so that we can collect a set of changes and release that as a version.  For now I see v0.0 as deprecated and v0.1 as current.

app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,7 @@ module V0
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
```
bdunne: This should be in the `V0x1::TasksController`, right?

eclarizio: Gah! Yes, you are correct.

Issue #123, opened 2019-02-12
Can't resolve './ID'


Hyperkid123:
When trying to build API client from generated code i get following error:

```
ERROR in ./src/model/IDReadOnly.js
Module not found: Error: Can't resolve './ID' in '/home/mmarosi/service-portal/topological_inventory-api-jsclient-1/src/model'
 @ ./src/model/IDReadOnly.js 20:0-22 39:4-6 65:8-10 66:8-10
 @ ./src/model/Authentication.js
 @ ./src/index.js
error Command failed with exit code 2.
```

I am using API code gen and latest v0.1.0 schema. When i check the generated source code there is truly no `ID.js` file. Last successful client generation was from https://github.com/ManageIQ/topological_inventory-api-jsclient/commit/e61a221537fdade2fb3f7270e4b10091d910a371 (Feb 6).

When i was going through the API commits i think this might have something to do with it: https://github.com/ManageIQ/topological_inventory-api/commit/35196a77c7d4881c2a3e5b998b172cab7f832536

cc @carbonin @bdunne

martinpovolny:
In the previous iteration I was fixing this by removing any references to the ID from the generated javascripts. But that was with the previous generator (swagger, not openAPI).

Hyperkid123:
It did work with the OpenAPI a week. On version 0.1

bdunne:
@Hyperkid123 If you rename `parameters/ID` to something else, does that work for you?

The validator says there's nothing wrong with the documents on master...
```
$ swagger-cli validate public/doc/swagger-2-v0.1.0.yaml 
public/doc/swagger-2-v0.1.0.yaml is valid

$ swagger-cli validate public/doc/swagger-2-v0.0.1.yaml 
public/doc/swagger-2-v0.0.1.yaml is valid
```

Hyperkid123:
@bdunne so if u rename `parameters/ID` anywhere in the yam i get an error

```
Error resolving $ref pointer "/home/mmarosi/service-portal/topological_inventory-api/public/doc/swagger-2-v0.1.0.yaml#/parameters/FOO". 
Token "FOO" does not exist.
```

Otherwise its valid. Or do you have something else in mind?

bdunne:
> so if i rename `parameters/ID` anywhere in the yam i get an error

I meant rename all references in the YAML from `parameters/ID` to `parameters/Foo`

Hyperkid123:
[...]

PR #139, opened 2019-02-28
Fix ID and IDReadonly.

martinpovolny:
The code generators do not expect models to be non-objects.

Fixes https://github.com/ManageIQ/topological_inventory-api/issues/123

might also fix https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/3

Ping @agrare, @bdunne, @eclarizio, @Hyperkid123 

Generating the javascript client from fixed swagger file w/o any issues and hand-editing:

```
lucifer - [~/Projects/sources/topological_inventory-api-jsclient] (master)$ npm run build

> @manageiq/topological_inventory@1.0.6 build /home/martin/Projects/sources/topological_inventory-api-jsclient
> webpack --config ./conf/webpack.config.js

Hash: 6a51c643a4d9b97b7a05
Version: webpack 4.28.1
Time: 1618ms
Built at: 2019-02-28 14:09:54
  Asset     Size  Chunks             Chunk Names
main.js  608 KiB    main  [emitted]  main
Entrypoint main = main.js
[./src/ApiClient.js] 20.1 KiB {main} [built]
[./src/api/DefaultApi.js] 161 KiB {main} [built]
[./src/index.js] 10.4 KiB {main} [built]
[./src/model/Authentication.js] 4.63 KiB {main} [built]
[./src/model/AuthenticationsCollection.js] 3.46 KiB {main} [built]
[./src/model/AvailabilitiesCollection.js] 3.43 KiB {main} [built]
[./src/model/Availability.js] 4.66 KiB {main} [built]
[./src/model/CollectionLinks.js] 3.24 KiB {main} [built]
[./src/model/CollectionMetadata.js] 2.68 KiB {main} [built]
[./src/model/Container.js] 4.85 KiB {main} [built]
[./src/model/ContainerGroup.js] 6.01 KiB {main} [built]
[./src/model/ContainerGroupsCollection.js] 3.46 KiB {main} [built]
[./src/model/ContainerImage.js] 5.44 KiB {main} [built]
[./src/model/ContainerImagesCollection.js] 3.46 KiB {main} [built]
[./src/model/ContainerNode.js] 6.11 KiB {main} [built]
    + 55 hidden modules
```

I believe the issues with the Ruby client will also be fixed by this change.

### DIFF
--- a/public/doc/swagger-2-v0.1.0.yaml
+++ b/public/doc/swagger-2-v0.1.0.yaml
@@ -1406,14 +1406,17 @@ definitions:
         example: openshift_default
         type: string
       id:
-        "$ref": "#/definitions/IDReadOnly"
+        type: string
+        pattern: "/^\\d+$/"
+        readOnly: true
       name:
         example: OpenShift default
         type: string
       password:
         type: string
       resource_id:
-        "$ref": "#/definitions/ID"
+        type: string
+        pattern: "/^\\d+$/"
       resource_type:
         example: Endpoint
         type: string
@@ -1423,7 +1426,8 @@ definitions:
       status_details:
         type: string
       tenant_id:
-        "$ref": "#/definitions/ID"
+        type: string
+        pattern: "/^\\d+$/"
       username:
         example: user@example.com
         type: string
@@ -1462,7 +1466,9 @@ definitions:
         readOnly: true
         type: string
       id:
-        "$ref": "#/definitions/IDReadOnly"
+        type: string
+        pattern: "/^\\d+$/"
+        readOnly: true
       identifier:
         example: "/service_offerings"
         readOnly: true
@@ -1474,7 +1480,9 @@ definitions:
         format: date-time
         type: string
       resource_id:
-        "$ref": "#/definitions/IDReadOnly"
+        type: string
+        pattern: "/^\\d+$/"
+        readOnly: true
       resource_type:
         readOnly: true
         type: string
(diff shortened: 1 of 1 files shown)


Ladas:
@martinpovolny could you update the generator like @bdunne has done here https://github.com/ManageIQ/topological_inventory-api/pull/140? If we are removing also the `ID`. What @bdunne is changing there is the only code that generates the `ID` related stuff, so we can change that to naything.

Then generate the swagger.yaml with `rake swagger:generate`

eclarizio:
As @bdunne mentioned on standup I had a similar problem with https://github.com/ManageIQ/topological_inventory-api/pull/131, but that was because of the way IDReadOnly was nested and the generator failing to figure out how to process that. We determined that IDs should all just be read only anyway, so my question is does the javascript generator still have problems figuring it out if we just use the `"$ref": "#/definitions/ID"` with the readOnly property set? It would really suck if we had to duplicate the regex and other properties everywhere we need an ID.

public/doc/swagger-2-v0.1.0.yaml
```diff
@@ -1406,14 +1406,17 @@ definitions:
         example: openshift_default
         type: string
       id:
-        "$ref": "#/definitions/IDReadOnly"
+        type: string
+        pattern: "/^\\d+$/"
```
bdunne: I'd prefer not to duplicate this everywhere, is the issue that `IDReadOnly` depended on `ID`?  (that sounds like a bug in the code generator)
Hopefully this problem is resolved by #140

martinpovolny:
Just retested with the swagger file from #140 and seems the problem is really fixed :+1: 

So I can close this one.

I will update https://github.com/ManageIQ/topological_inventory-api-jsclient/pull/5 (well maybe the files will get generated the same, will check).

Anyway seems we can close https://github.com/ManageIQ/topological_inventory-api/issues/123 now.

Hyperkid123:
I did test it right now and with #140 the code gets generated and the client builds correctly.

martinpovolny:
Also the generated javascript source is the same except for the "ID of the resource" comment. So we should be good!

agrare:
Okay awesome I'll regenerate the topological_inventory-api-client-ruby to fix https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/3

PR #141, opened 2019-02-28
Add columns needed for host inventory sync

Ladas:


public/doc/swagger-2-v0.1.0.yaml
```diff
@@ -2534,6 +2537,9 @@ definitions:
         format: date-time
         readOnly: true
         type: string
+      mac_addresses:
+        readOnly: true
+        type: string
```
agrare: @Ladas should this be an array of strings?

Ladas: hm, yes it should

Ladas: hum I need to change the generator

PR #142, opened 2019-02-28
Add tags as subcollection

Ladas:
Adding tags as subcollection. For this we need a generic code for subcollections done with `:through` mapping table

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -18,12 +18,12 @@ Rails.application.routes.draw do
         resources :containers, :only => [:index]
         resources :tags,       :only => [:index]
       end
+      resources :container_images,        :only => [:index, :show] do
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -16,16 +16,23 @@ def put(*_args); end
       resources :containers,              :only => [:index, :show]
       resources :container_groups,        :only => [:index, :show] do
         resources :containers, :only => [:index]
+        resources :tags, :only => [:index]
       end
       resources :container_nodes,         :only => [:index, :show] do
         resources :container_groups, :only => [:index]
+        resources :tags, :only => [:index]
+      end
+      resources :container_images,        :only => [:index, :show] do
```
bdunne: Alphabetically this should be above `container_nodes`

PR #143, opened 2019-03-01
Add specs for index and subcollections for every route

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-api/pull/142
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/111


agrare:
Not sure why hakiri says it is red, looks like no warnings to me when I click through

Ladas:
@agrare yeah, probably some error in the evaluation. It should be unrelated to this

PR #144, opened 2019-03-05
Tasks need to be updatable through the API

agrare:
The openshift-operations worker is updating tasks through the API, so we
need to include the update mixin in the tasks controller.

### DIFF
--- a/app/controllers/api/v0/tasks_controller.rb
+++ b/app/controllers/api/v0/tasks_controller.rb
@@ -3,6 +3,7 @@ module Api
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
     end
   end
 end


app/controllers/api/v0/tasks_controller.rb
```diff
@@ -3,6 +3,7 @@ module V0
     class TasksController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
+      include Api::V0::Mixins::UpdateMixin
```
bdunne: I think this should only be in v0.1 only.  Also it requires a change to the swagger.yaml

agrare: Okay, and it is already in the swagger.yaml: https://github.com/ManageIQ/topological_inventory-api/blob/master/public/doc/swagger-2-v0.1.0.yaml#L1407-L1427

bdunne: It's already in the V0x1::TasksController too: https://github.com/ManageIQ/topological_inventory-api/blob/master/app/controllers/api/v0x1.rb#L83

agrare: Hm I must have been using v0.0 when it was failing before, v0.1 already has the mixin

agrare:
Ran into this when testing the nil id issue so I think I went back to 0.0 as a test and hit this.  Works fine now so going to close this

PR #147, opened 2019-03-06
Fix tests failing due to changed metadata in paginated responses

bdunne:
https://github.com/ManageIQ/manageiq-api-common/pull/35
This should probably move to manageiq-api-common (and be a real matcher), but let's get things green first


martinpovolny:
Oh, I could have caught this. I tested the PR with this API, but did not run the tests :-(

PR #150, opened 2019-03-11
Send message to Kafka when a task is updated

eclarizio:
I'm not sure if the payload may need to change, depending on what other data catalog needs.

Also the message service I just made up a name, do we have a better name than "platform.topological-inventory.events"?

@agrare Please review 😄

### DIFF
--- a/app/controllers/api/v0x1/tasks_controller.rb
+++ b/app/controllers/api/v0x1/tasks_controller.rb
@@ -1,0 +1,21 @@
+module Api
+  module V0x1
+    class TasksController < Api::V0x0::TasksController
+      include Api::V0x1::Mixins::IndexMixin
+
+      def update
+        model.update(params.require(:id), params_for_update)
+
+        messaging_client.publish_message(
+          :service => "platform.topological-inventory.task-output-stream",
+          :message => "Task.update",
+          :payload => params_for_update.to_h.merge("task_id" => params.require(:id))
+        )
+
+        head :no_content
+      rescue ActiveRecord::RecordNotFound
+        head :not_found
+      end
+    end
+  end
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -144,6 +144,17 @@ class ApplicationController < ActionController::API
     self.class.send(:api_doc_definition)
   end
 
+  def messaging_client
+    require "manageiq-messaging"
+
+    @messaging_client ||= ManageIQ::Messaging::Client.open({
+      :protocol => :Kafka,
+      :host     => ENV["QUEUE_HOST"] || "localhost",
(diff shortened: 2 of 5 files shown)


app/controllers/api/v0x1/tasks_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Api
+  module V0x1
+    class TasksController < Api::V0x0::TasksController
+      include Api::V0x1::Mixins::IndexMixin
+
+      def update
+        model.update(params.require(:id), params_for_update)
+
+        messaging_client.publish_message(
+          :service => "platform.topological-inventory.events",
```
agrare: Since this is going to be specific to task updates (similar to how the persister emits inventory updates) we can be consistent with the naming of that topic so how about `platform.topological-inventory.task-output-stream`?

app/controllers/application_controller.rb
```diff
@@ -144,6 +144,21 @@ def api_doc_definition
     self.class.send(:api_doc_definition)
   end
 
+  def messaging_client
+    require "manageiq-messaging"
+
+    @messaging_client ||= ManageIQ::Messaging::Client.open(messaging_opts)
+  end
+
+  def messaging_opts
```
agrare: I know this was just a copy & paste but since these are constant and only ever used by https://github.com/ManageIQ/topological_inventory-api/pull/150/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR150 can you just put this in the `messaging_client` method like
```
@messaging_client ||= ManageIQ::Messaging::Client.open({
  :protocol => :kafka,...
})
```

app/controllers/api/v0x1/tasks_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Api
+  module V0x1
+    class TasksController < Api::V0x0::TasksController
```
agrare: Could you add this to the `V0` base `TasksController` instead of just to `v0x1`?  We might need to update the swagger for `v0.0.1` to add patch for tasks but that shouldn't be that bad.

eclarizio: Brandon didn't seem to want this when I made the original update PR #137. Specifically [this comment](https://github.com/ManageIQ/topological_inventory-api/pull/137#discussion_r260426103) makes me think we don't want to make any more changes to `v0.0.1`.

app/controllers/application_controller.rb
eclarizio: 👍

app/controllers/api/v0x1/tasks_controller.rb
agrare: Hm okay, well we're going to deprecate that soon anyway so we'll just have to move this from v0x1 to v0 then

PR #151, opened 2019-03-11
Fix for TPINVTRY-224 - Inconsistent swagger response codes for resource updates

abellotti:
This fixes https://projects.engineering.redhat.com/browse/TPINVTRY-224

The 400 is actually the proper return code for trying to update read-only attributes,
i.e. a Bad Request. We generate the 400's properly as we catch
the ActionController::UnpermittedParameters error.

Fix was to update the swagger_generate.rake script to include a 400 response
code for collection updates.  This is effectively reflected for
the authentications, endpoints, sources and tasks collections.


agrare:
Good to have you back @abellotti :D

PR #156, opened 2019-03-14
API Filtering

bdunne:
- Add some basic filtering operations for strings, integers and datetimes.  
- Allow for bulk query when the `eq` comparators are used with an array of values.

cc @abellotti

### DIFF
--- a/app/controllers/application_controller/filter.rb
+++ b/app/controllers/application_controller/filter.rb
@@ -14,13 +14,13 @@ class ApplicationController
     def apply
       return query if @raw_filter.blank?
       @raw_filter.each do |k, v|
-        with_attribute_for_key(k) do |attribute|
-          if attribute["type"] == "string"
-            type = determine_string_attribute_type(attribute)
-            send(type, k, v)
-          else
-            errors.add("unsupported attribute type for: #{k}")
-          end
+        next unless attribute = attribute_for_key(k)
+
+        if attribute["type"] == "string"
+          type = determine_string_attribute_type(attribute)
+          send(type, k, v)
+        else
+          errors.add("unsupported attribute type for: #{k}")
         end
       end
 
@@ -40,9 +40,11 @@ class ApplicationController
       end
     end
 
-    def with_attribute_for_key(key)
+    def attribute_for_key(key)
       attribute = api_doc_definition.properties[key.to_s]
-      attribute ? yield(attribute) : errors.add("found unpermitted parameter: #{key}")
+      return attribute if attribute
+      errors.add("found unpermitted parameter: #{key}")
+      nil
     end
 
     def determine_string_attribute_type(attribute)
@@ -66,7 +68,8 @@ class ApplicationController
     end
 
     def add_filter(requested_comparator, allowed_comparators, key, value)
-      type = with_attribute_for_key(key) { |attribute| determine_string_attribute_type(attribute) }
+      return unless attribute = attribute_for_key(key)
+      type = determine_string_attribute_type(attribute)
 
       if requested_comparator.in?(["nil", "not_nil"])
         send("comparator_#{requested_comparator}", key, value)
--- a/app/controllers/application_controller/filter.rb
+++ b/app/controllers/application_controller/filter.rb
@@ -1,0 +1,152 @@
+class ApplicationController
+  class Filter
+    INTEGER_COMPARISON_KEYWORDS = ["eq", "gt", "gte", "lt", "lte", "nil", "not_nil"].freeze
+    STRING_COMPARISON_KEYWORDS  = ["contains", "eq", "starts_with", "ends_with", "nil", "not_nil"].freeze
+
+    attr_reader :apply, :api_doc_definition
+
+    def initialize(model, raw_filter, api_doc_definition)
+      self.query          = model
+      @raw_filter         = raw_filter
+      @api_doc_definition = api_doc_definition
+    end
+
+    def apply
+      return query if @raw_filter.blank?
+      @raw_filter.each do |k, v|
+        with_attribute_for_key(k) do |attribute|
+          if attribute["type"] == "string"
+            type = determine_string_attribute_type(attribute)
+            send(type, k, v)
+          else
+            errors.add("unsupported attribute type for: #{k}")
+          end
+        end
+      end
+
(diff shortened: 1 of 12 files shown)


bdunne:
I was trying to finish this up last Friday, but didn't quite get there.  I think I have most of the test cases covered, but you may want to double check that.  IIRC, I was last trying to deduplicate some of the logic in the `Filter` class and make it look a little bit better, but it should all be working.

app/controllers/application_controller/filter.rb
```diff
@@ -0,0 +1,152 @@
...
+          add_filter(comparator, INTEGER_COMPARISON_KEYWORDS, k, value)
+        end
+      else
+        add_filter("eq", INTEGER_COMPARISON_KEYWORDS, k, val)
+      end
+    end
+
+    def comparator_contains(key, value)
+      return value.each { |v| comparator_contains(key, v) } if value.kind_of?(Array)
+      self.query = query.where("#{key} LIKE ?", "%#{value}%")
```
Fryguy: we should probably quote the key and elsewhere...either that or use `?` (though I'm not sure you *can* use `?` in the key part.  I'm concerned about possible injection with this via query params.

app/controllers/application_controller/filter.rb
```diff
@@ -0,0 +1,152 @@
...
+      attr_reader :error_document
+
+      def initialize(attrs)
+        @error_document = attrs[:error_document]
+      end
+    end
+
+    def with_attribute_for_key(key)
+      attribute = api_doc_definition.properties[key.to_s]
+      attribute ? yield(attribute) : errors.add("found unpermitted parameter: #{key}")
```
Fryguy: If you have an error here, then it just skips the yield?  If so, won't that allow add_filter to continue on?

Fryguy: I'm curious why this is in yield-style?  Seems just as easy for this method to return the value and let the caller do whatever it needs, since there's no "cleanup" afterwards that a block would be needed for.

spec/requests/application_controller/filter_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    it("single with 'eq' key")                    { expect_success("filter[id][eq]=#{task_1.id}", task_1) }
+    it("array of values with 'eq' key")           { expect_success("filter[id][eq][]=#{task_1.id}&filter[id][eq][]=#{task_2.id}", task_1, task_2) }
+
+    it("key:gt")                                  { expect_success("filter[id][gt]=#{task_2.id}", task_3, task_4) }
+    it("key:gte")                                 { expect_success("filter[id][gte]=#{task_2.id}", task_2, task_3, task_4) }
+    it("key:lt")                                  { expect_success("filter[id][lt]=#{task_3.id}", task_1, task_2) }
+    it("key:lte")                                 { expect_success("filter[id][lte]=#{task_3.id}", task_1, task_2, task_3) }
+  end
+
+  context "timestamps" do
```
Fryguy: do we support gt, gte, etc for timestamps?

spec/requests/application_controller/filter_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    it "unsupported attribute type" do
+      source_type = SourceType.create!(:name => "a", :vendor => "a", :product_name => "a")
+      source = Source.create!(:source_type => source_type, :tenant => tenant, :name => "a")
+      Vm.create!(:source => source, :tenant => tenant, :source_ref => "a")
+
+      get("/api/v0.1/vms?filter[mac_addresses]=a")
+
+      expect(response.status).to eq(400)
+      expect(response.parsed_body["errors"]).to eq([{"detail"=>"unsupported attribute type for: mac_addresses", "status"=>400}])
```
Fryguy: We might want a spec for multiple bad keys, or perhaps a mix of good and bad keys.

Fryguy:
This is awesome :)

spec/requests/application_controller/filter_spec.rb
bdunne: Yes , all of the Integer comparisons listed here.  https://github.com/ManageIQ/topological_inventory-api/pull/156/files#diff-08a3013ef490c3273c6b0ffc7c84a0eaR3

spec/requests/application_controller/filter_spec.rb
bdunne: I agree, that would be great.

app/controllers/application_controller/filter.rb
bdunne: Yeah, this is one of the things I didn't get to cleaning up.

app/controllers/application_controller/filter.rb
bdunne: I think the `key` is safe here since `with_attribute_for_key` will error out if the exact key is not found in the swagger doc.

bdunne: Oh, I see, that should be fixed now.

app/controllers/application_controller/filter.rb
bdunne: Fixed.

bdunne:
@abellotti Can you add the remaining tests that @Fryguy mentioned?  I'm not sure if you guys want to do that in this PR or another.

bdunne:
Also TODO: The swagger generator and v0.1 doc need to be updated to include the `filter` query param for all collections.

abellotti:
@bdunne this looks great !! minor thing above.

nice & clean interface !! we can always extend later to support multi-values for comparators like starts_with and the like.

yea, let's get that in, I can write the additional tests in a follow-up PR. 

re: the TODO for the spec, we might need to go to openapi 3.0 (query parameters type deepObject for things like filter[] ...).  so merge this, PR for additional specs, PR to move to OpenAPI Spec 3.0, PR to add the filter[] specification using OpenAPI 3.0.

abellotti:
This PR should be closed as the commits are moved to https://github.com/ManageIQ/topological_inventory-api/pull/161 and additional changes going there.

Fryguy:
@abellotti As long as you haven't modified @bdunne 's commits, then we it being open or not will be handled automatically.

Fryguy:
Closed via #156

PR #157, opened 2019-03-14
Enhancement to move to OpenAPI Spec 3.0.0

abellotti:
Enhancement to update the API to:
- move from swagger 2 to openapi 3.0.0
- rename the swagger:generate task to  openapi:generate and update accordingly.

Updating OpenAPI 3.0 specs now done via:
  $ bundle exec rake openapi:generate


abellotti:
Marking as WIP, holding place for this work until we're ready to pull the trigger.

lib/api/docs.rb
```diff
@@ -1,3 +1,4 @@
 module Api
-  Docs = ::OpenApi::Docs.new(Dir.glob(Rails.root.join("public/doc/swagger*.yaml")))
+  Docs = ::OpenApi::Docs.new(Dir.glob(Rails.root.join("public/doc/swagger*.yaml")) +
+                             Dir.glob(Rails.root.join("public/doc/openapi*.yaml")))
```
Fryguy: In catalog, and in the IPP, I believe we want to go to .json format for the new openapi file.

lib/tasks/openapi_generate.rake
```diff
@@ -0,0 +1,388 @@
...
+      r = ActionDispatch::Routing::RouteWrapper.new(route)
+      next if r.internal? # Don't display rails routes
+      next if r.engine? # Don't care right now...
+
+      array << r
+    end
+  end
+
+  def openapi_file
+    Pathname.new(__dir__).join("../../public/doc/openapi-3-v0.2.0.yaml").to_s
```
Fryguy: This feels weird that we would hardcode the 0.2.0 version in the generator...I'd expect it as a parameter or something.

lib/tasks/openapi_generate.rake
```diff
@@ -0,0 +1,388 @@
+class OpenapiGenerator
```
Fryguy: I see most of this is just a copy of the existing swagger generator.  I'm concerned about duplication, but that can be cleaned up in a follow up.  Even so, can you show a diff between the two so I can figure out how to review this?

lib/api/docs.rb
abellotti: yep, cleaner too. moved to openapi*.json for the OpenAPI 3.x specifications.

lib/tasks/openapi_generate.rake
abellotti: I think this is now better, I've essentially git mv'd the original swagger_generate.rake to the new openapi_generate.rake successor, so git diffs now show the changes.

lib/tasks/openapi_generate.rake
abellotti: Agreed. version is now generated from the latest openapi.json route defined for the project.

lib/api/docs.rb
```diff
@@ -1,3 +1,4 @@
 module Api
-  Docs = ::OpenApi::Docs.new(Dir.glob(Rails.root.join("public/doc/swagger*.yaml")))
+  Docs = ::OpenApi::Docs.new(Dir.glob(Rails.root.join("public", "doc", "openapi*.json")) +
+                             Dir.glob(Rails.root.join("public", "doc", "swagger*.yaml")))
```
Fryguy: Can we just drop the yamls?

Fryguy: Or put another way, can we just drop v2 altogether, since we are going to cut a 1.0 anyway

app/controllers/api/v0x1.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module V0x1
     class RootController < ApplicationController
       def openapi
-        render :json => Api::Docs["0.1"].to_json
+        render :json => Api::Docs["0.1"].content.to_json
```
Fryguy: This change is not clear to me.  Why doesn't the to_json itself just call content?

lib/open_api/docs/component_collection.rb
```diff
@@ -13,7 +13,7 @@ def [](name)
       end
 
       def load_definition(name)
-        raw_definition = @doc.content[@category][name]
+        raw_definition = @doc.content.fetch_path(@category.split("/"))[name]
```
Fryguy: If you're using fetch_path, you can put the name in the fetch as well

```ruby
@doc.content.fetch_path(*@category.split("/"), name)
```

lib/api/docs.rb
abellotti: We can drop the yamls & v2 Doc library since we don't expose /api/v0.0/openapi.json anyways,  want me to do in this PR ?

lib/open_api/docs.rb
```diff
@@ -7,7 +7,11 @@ def initialize(glob)
 
     def load_file(file)
       yaml = YAML.load_file(file)
-      doc  = DocV2.new(yaml) if yaml["swagger"] == "2.0"
+      doc = if yaml["swagger"] == "2.0"
+              DocV2.new(yaml)
+            elsif yaml["openapi"] =~ /3*/
+              DocV3.new(yaml)
+            end
```
Fryguy: Not sure I follow the need for this if we don't have v2 anymore.  Additionally, it's not a yaml anymore, right, so is the variable name wrong?

abellotti: yaml really just a hash :) this code would change of course if we drop DocV2 altogether.

lib/api/docs.rb
Fryguy: yup 👍

app/controllers/api/v0x1.rb
abellotti: it does, it's just that the spec should not include the content key which is what to_json serializes it as:

```
{
  "content" : {
    "openapi" : "3.0.0",
    "info" : { ... },
    "security" : { ... },
    "paths" : { 
       ...
    }
    ...
}
```

should just be:

```
{
  "openapi" : "3.0.0",
  "info" : { ... },
  "security" : { ... },
  "paths" : {
     ...
  }
  ...
}
```

same issue with the current v0.1.0 swagger flavor.

lib/open_api/docs/component_collection.rb
abellotti: 👍

abellotti:
Ok @Fryguy, I've removed the DocV2 code and simplified what we had in this PR.  I've also converted the v0.0.1 yaml to OpenAPI 3.0.0 json as this would be needed for the /internal/v0 paths anyways.  It also allows us to remove /api/v0.0 in a separate PR. let me know if you'd like to see other changes in this PR. Thank you.

lib/tasks/openapi_generate.rake
```diff
@@ -0,0 +1,416 @@
...
+      when :datetime
+        properties_value["format"] = "date-time"
+      when :integer
+        properties_value["type"] = "integer"
+      when :float
+        properties_value["type"] = "number"
+      when :boolean
+        properties_value["type"] = "boolean"
+      when :jsonb
+        ['type', 'items'].each do |property_key|
```
Ladas: hmm you're missing this line https://github.com/ManageIQ/topological_inventory-api/blob/b9c30161c95b8fef9d85a002674ab04b7d796c3c/lib/tasks/swagger_generate.rake#L177

it was probably added after you took the generator?

abellotti: 👍  Good catch!!  Thanks @Ladas for 👀 at this.  Will push an update.

lib/open_api/docs.rb
```diff
@@ -6,8 +6,8 @@ def initialize(glob)
     end
 
     def load_file(file)
-      yaml = YAML.load_file(file)
-      doc  = DocV2.new(yaml) if yaml["swagger"] == "2.0"
+      openapi_spec = JSON.parse(File.read(file))
+      doc = DocV3.new(openapi_spec) if openapi_spec["openapi"] =~ /3*/
```
Fryguy: if it doesn't match `/3*/` what happens?  Should it raise?

lib/open_api/docs/doc_v3.rb
```diff
@@ -11,37 +11,34 @@ def version
...
       def example_attributes(key)
-        definitions[key]["properties"].each_with_object({}) do |(col, stuff), hash|
+        schemas[key]["properties"].each_with_object({}) do |(col, stuff), hash|
           hash[col] = stuff["example"] if stuff.key?("example")
         end
       end
 
       def base_path
-        @base_path ||= @content["basePath"]
+        @base_path ||= @content["servers"].first["variables"]["basePath"]["default"]
```
Fryguy: Minor...

```ruby
@content.fetch_path("servers, 0, "variables", "basePath", "default")
```

lib/open_api/docs/doc_v3.rb
```diff
@@ -11,37 +11,34 @@ def version
...
       def routes
-        @routes ||= begin
-          paths.flat_map do |path, hash|
-            hash.collect do |verb, _details|
-              p = File.join(base_path, path).gsub(/{\w*}/, ":id")
-              {:path => p, :verb => verb.upcase}
-            end
-          end
-        end
+        @routes ||= Docs.path_routes(base_path, paths)
```
Fryguy: I don't think you need this refactor anymore...you did before when you were trying to share v2 and v3 code, but now you don't.

abellotti:
Ping @Fryguy, all changes made and pushed out. Anything else needed here ?

mkanoor:
@abellotti 
Line 782
      operationId: showServiceOfferingIcon
Line 798
      operationId: showServiceOfferingIcon
They have to be unique, when you load the son file in the swagger editor it complains

abellotti:
I know, let's fix that after this PR is merged since it's a current issue with the swagger too.  Essentially the openapi.generate needs to be enhanced to know about the case where we're getting a single subresource (and not listing), so the operationId generated for this would be something like: "showServiceOfferingIconIconData". Thanks.

PR #158, opened 2019-03-20
Changing the default base prefix from /r/insights/platform to /api

abellotti:
- updated the swagger specs to /api since that's where we're now deployed


abellotti:
@mkanoor   I think this what you were looking for.

@agrare or @gtanzillo can I borrow your 👀 for this ? Thanks.

mkanoor:
👍 Please merge

PR #159, opened 2019-03-21
Add route for service offering icon with right image content type

Hyperkid123:
With this change we can serve images directly and use classic image tag on client.
```
<img src="http://localhost:3000/api/v0.1/service_offering_icons/2/icon_data" />
```

Screenshot from 2019-03-21 17-51-39

cc @skateman

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'json-schema',        '~> 2.8'
 gem 'manageiq-loggers',   '~> 0.1'
 gem 'manageiq-messaging', '~> 0.1.2', :require => false
 gem 'manageiq-password',  '~> 0.2', ">= 0.2.1"
+gem 'mimemagic',          '~> 0.3.3'
 gem 'pg',                 '~> 1.0', :require => false
 gem 'puma',               '~> 3.0'
 gem 'rack-cors',          '>= 0.4.1'
--- a/app/controllers/api/v0/service_offering_icons_controller.rb
+++ b/app/controllers/api/v0/service_offering_icons_controller.rb
@@ -4,6 +4,16 @@ module Api
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
 
+      def icon_data
+        service_offering_icon = model.find(params[:service_offering_icon_id].to_i)
+
+        send_data(service_offering_icon.data, 
+          :type         => MimeMagic.by_magic(service_offering_icon.data).type,
+          :disposition  => 'inline')
+      rescue ActiveRecord::RecordNotFound
+        head :not_found
+      end
+
       private
 
       def model
--- a/spec/requests/api/v0.1/service_offering_icon_spec.rb
+++ b/spec/requests/api/v0.1/service_offering_icon_spec.rb
@@ -4,12 +4,18 @@ RSpec.describe("v0.1 - ServiceOfferingIcon") do
   let(:source) { Source.create!(:name => "name", :source_type => source_type, :tenant => tenant) }
   let(:source_type) { SourceType.create!(:vendor => "vendor", :product_name => "product_name", :name => "name") }
   let(:tenant) { Tenant.create! }
+  let(:data) { '<svg width=\"580\" height=\"400\" xmlns=\"http://www.w3.org/2000/svg\"> <!-- Created with Method Draw -'\
+    'http://github.com/duopixel/Method-Draw/ --> <g>  <title>background</title>  <rect fill=\"#ff0000\" '\
+    'id=\"canvas_background\" height=\"402\" width=\"582\" y=\"-1\" x=\"-1\"/>  <g display=\"none\" overflow=\"visible\"'\
+    ' y=\"0\" x=\"0\" height=\"100%\" width=\"100%\" id=\"canvasGrid\">   <rect fill=\"url(#gridpattern)\" '\
+    'stroke-width=\"0\" y=\"0\" x=\"0\" height=\"100%\" width=\"100%\"/>  </g> </g> <g>  <title>Layer 1</title> </g></svg>' }
 
   let(:attributes) do
     {
       "source_id"            => source.id.to_s,
       "tenant_id"            => tenant.id.to_s,
-      "source_ref"           => SecureRandom.uuid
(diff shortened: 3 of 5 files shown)


app/controllers/api/v0/service_plans_controller.rb
```diff
@@ -16,7 +16,7 @@ def order
 
         render :json => {:task_id => task.id}
       rescue ActiveRecord::RecordNotFound
-        head :bad_request
+        head :not_found
```
Ladas: @Hyperkid123 I think we should change this back, then the CI will pass.

@bdunne does the IPP says we should return 400 for not found btw?

app/controllers/api/v0/service_offering_icons_controller.rb
```diff
@@ -4,6 +4,15 @@ class ServiceOfferingIconsController < ApplicationController
       include Api::V0::Mixins::IndexMixin
       include Api::V0::Mixins::ShowMixin
 
+      def icon_data
+        service_offering_icon = model.find(params[:service_offering_icon_id].to_i)
+
+        send_data(service_offering_icon.data, 
+          :type         => MimeMagic.by_magic(service_offering_icon.data).type,
```
Ladas: @agrare @bdunne are we ok with new gem for recognizing the content type?

app/controllers/api/v0/service_plans_controller.rb
Hyperkid123: oh right forgot to discard that change

app/controllers/api/v0/service_offering_icons_controller.rb
Fryguy: Seems ok to me.

app/controllers/api/v0/service_offering_icons_controller.rb
```diff
@@ -4,6 +4,15 @@ class ServiceOfferingIconsController < ApplicationController
...
+          :disposition  => 'inline')
+      rescue ActiveRecord::RecordNotFound
+        head :not_found
```
Fryguy: Can we just let this blow all the way out instead of rescuing?

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'pg',                 '~> 1.0', :require => false
 gem 'puma',               '~> 3.0'
 gem 'rack-cors',          '>= 0.4.1'
 gem 'rails',              '>= 5.2.2.1', '~> 5.2.2'
+gem 'mimemagic',          '~> 0.3.3'
```
Fryguy: Minor, but alphabetize it in the list.

app/controllers/api/v0/service_offering_icons_controller.rb
```diff
@@ -4,6 +4,15 @@ class ServiceOfferingIconsController < ApplicationController
...
+      end
       private
```
agrare: There should be a newline between the last method's `end` and `private`

spec/requests/api/v0.1/service_offering_icon_spec.rb
```diff
@@ -18,4 +24,49 @@
...
+
+    context "get" do
+      it "success: with a valid id" do
+        instance = primary_collection.to_s.singularize.camelize.constantize.create!(attributes)
+
+        get(subcollection_path(instance.id))
+
+        expect(response).to have_attributes(
+                              :status      => 200,
+                              :parsed_body => data
```
agrare: Can you test that the content type is set properly?

app/controllers/api/v0/service_offering_icons_controller.rb
agrare: This is interesting actually, not sure if it is intentional or not but all of the current index methods do this for the primary collect, but the subcollection uses the `rescue_from` error document from the base ApplicationController.

Example index spec for the primary collection: https://github.com/ManageIQ/topological_inventory-api/blob/master/spec/requests/api/v0.1/shared_examples_for_index.rb#L45-L54
Example index spec for the subcollection: https://github.com/ManageIQ/topological_inventory-api/blob/master/spec/requests/api/v0.1/shared_examples_for_index.rb#L88-L99

@abellotti do you know if this is intentional?  It would seem that if we remove all of these it would be more consistent.

Ladas: At least we should start to be consistent with the swagger file, which says we should return 404, not 400.

Ladas: hm, ok we return 400 only in ServicePlanController, the rest is ok

spec/requests/api/v0.1/service_offering_icon_spec.rb
```diff
@@ -18,4 +24,50 @@
...
+    context "get" do
+      it "success: with a valid id" do
+        instance = primary_collection.to_s.singularize.camelize.constantize.create!(attributes)
+
+        get(subcollection_path(instance.id))
+
+        expect(response).to have_attributes(
+                              :status       => 200,
+                              :parsed_body  => data,
+                              :content_type => "image/svg+xml"
```
agrare: :+1: nice

agrare:
Hey awesome work @Hyperkid123 !

Hyperkid123:
Thanks

app/controllers/api/v0/service_offering_icons_controller.rb
abellotti: not sure of the intent/historical reason, but from what I can tell, if we nix the rescue's and just leverage the rescue_from in the app controller for the ActiveRecord::RecordNotFound errors, instead of getting an empty response body with the 404, you'd get the:

```
{
  "errors" : [
     { "details" => "Couldn't find <type> with 'id'=....,
        "status" => 404
      }
  ]
}
```

Which isn't entirely bad. The specs would just need updating for the different expectations, but like you said, more consistent.

agrare: :+1: @abellotti yeah I agree, I'll work on updating the specs

agrare: @abellotti https://github.com/ManageIQ/topological_inventory-api/pull/164

PR #161, opened 2019-03-21
API Filtering plus

abellotti:
Brandon's API Filtering PR https://github.com/ManageIQ/topological_inventory-api/pull/156 +
- quieting down hakiri - now using arel
- properly escaping SQL like patterns so we can also match characters that include %
- additional tests

### DIFF
--- a/app/controllers/application_controller/filter.rb
+++ b/app/controllers/application_controller/filter.rb
@@ -3,10 +3,11 @@ class ApplicationController
     INTEGER_COMPARISON_KEYWORDS = ["eq", "gt", "gte", "lt", "lte", "nil", "not_nil"].freeze
     STRING_COMPARISON_KEYWORDS  = ["contains", "eq", "starts_with", "ends_with", "nil", "not_nil"].freeze
 
-    attr_reader :apply, :api_doc_definition
+    attr_reader :apply, :arel_table, :api_doc_definition
 
     def initialize(model, raw_filter, api_doc_definition)
       self.query          = model
+      @arel_table         = model.arel_table
       @raw_filter         = raw_filter
       @api_doc_definition = api_doc_definition
     end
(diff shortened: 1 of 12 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,5 @@
+---
+version: '2'
+checks:
+  method-count:
+    enabled: false
```
Fryguy: This should not be necessary...we likely just missed adding the general .codeclimate.yml to this.

Fryguy: You can ignore codeclimate for this PR as their issues are probably too strict for us.

lib/topological_inventory/api/error_document.rb
```diff
@@ -1,19 +1,27 @@
 module TopologicalInventory
   module Api
-    class ErrorDocument < Hash
-      def initialize
-        self["errors"] = []
-      end
-
-      def add(status, message)
+    class ErrorDocument
```
Fryguy: Let's leave this the way it was for now...too much refactoring in a PR adding a new feature and it's cluttering up the diff.

Fryguy: Oh I see Brandon did this in his original commits...let's leave it then...I don't want to mess with his original commits.

.codeclimate.yml
abellotti: ok, reverted this.

lib/topological_inventory/api/error_document.rb
abellotti: ok, we can always refactor  later.

app/controllers/application_controller/filter.rb
```diff
@@ -0,0 +1,156 @@
...
+      else
+        add_filter("eq", STRING_COMPARISON_KEYWORDS, k, val)
+      end
+    end
+
+    def add_filter(requested_comparator, allowed_comparators, key, value)
+      return unless attribute = attribute_for_key(key)
+      type = determine_string_attribute_type(attribute)
+
+      if requested_comparator.in?(["nil", "not_nil"])
```
Fryguy: super minor, but we should extract these into a NIL_COMPARISON_KEYWORDS constants in a follow up PR.

bdunne:
@abellotti Are there any tests for the SQL injection cases?

PR #165, opened 2019-03-26
Enforce tenancy (second try)

gtanzillo:
Enforce tenancy by default. Tenancy can be bypassed by setting the environment variable `BYPASS_TENANCY` to "true"

This PR is a replacement of the [original PR](https://github.com/ManageIQ/topological_inventory-api/pull/104) which got out of sync with master due to moving and reworking on many tests. The main changes and tests that directly test tenancy have been pulled in from that PR. 

However, updates to the many existing tests to make them pass with enforcement on by default have been left behind because the in order to get something that can be merged soon.


app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
 
   private
 
+  set_current_tenant_through_filter
+  before_action :set_the_current_tenant
+  after_action  :unset_the_current_tenant
+
+  def set_the_current_tenant
+    return unless Tenant.tenancy_enabled?
```
Fryguy: If tenancy is disabled, then doesn't the sources_controller end up creating a tenant anyway since it executes before this line?

app/controllers/api/v0/sources_controller.rb
```diff
@@ -6,6 +6,13 @@ class SourcesController < ApplicationController
       include Api::V0::Mixins::ShowMixin
       include Api::V0::Mixins::UpdateMixin
 
+      def set_the_current_tenant
+        # Auto-create tenant if creating a new source and tenant does not exist
+        Tenant.find_or_create_by(:external_tenant => user_identity) if action_name == "create"
```
Fryguy: why is user_identity stored in the external_tenant?

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
...
+
+    tenant = Tenant.find_by(:external_tenant => user_identity)
+    if tenant
+      set_current_tenant(tenant)
+    else
+      error_document = {
+          "errors" => [
```
Fryguy: 2 space indent

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
...
   private
 
+  set_current_tenant_through_filter
+  before_action :set_the_current_tenant
+  after_action  :unset_the_current_tenant
+
+  def set_the_current_tenant
+    return unless Tenant.tenancy_enabled?
+
+    tenant = Tenant.find_by(:external_tenant => user_identity)
```
Fryguy: Can we just let this blow up naturally with an ActiveRecord::RecordNotFound and go through the normal path instead of building our own error document?

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
 
   private
 
+  set_current_tenant_through_filter
+  before_action :set_the_current_tenant
+  after_action  :unset_the_current_tenant
```
Fryguy: This might be better designed as an around_action since they are a joined pair.

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
...
+      }
+      render :json => error_document, :status => :unauthorized
+    end
+  end
+
+  def unset_the_current_tenant
+    set_current_tenant(nil)
+  end
+
+  def user_identity
```
Fryguy: Recommend renaming since it's not the user_identity, but the user's tenant.

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
...
+    ManageIQ::API::Common::Request.with_request(request) do |current|
```
Fryguy: This should probably be around the entire request with a before_action (though if that's breaking a lot of things we can do in a followup)

app/controllers/api/v0/sources_controller.rb
gtanzillo: It's the account number. I'll rename that method

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,39 @@ class ApplicationController < ActionController::API
...
+      current.user.tenant if current.headers.key?(ManageIQ::API::Common::Request::IDENTITY_KEY)
```
Fryguy: checking the key this way may not be case-insensitive, so this might not be a accurate.  Perhaps just get the current.user.tenant and catch the exception instead?  It should be set, so exception handling feels like the right path.  Additionally, if there is no tenant, what happens?  This code doesn't raise, so it just moves on?

app/controllers/application_controller.rb
gtanzillo: I tried to do it that way but it seemed to bypass Brandon's code that checks parameters so I backed it out. Let me try again and maybe we can look at it together

spec/support/tenant_identity.rb
```diff
@@ -0,0 +1,15 @@
...
+      included do
+        let!(:tenant)           { Tenant.create!(:external_tenant => external_tenant) }
+        let!(:external_tenant)  { rand(1000).to_s }
+        let!(:unknown_tenant)   { rand(1000).to_s }
+        let!(:identity)         { Base64.encode64({'identity' => { 'account_number' => external_tenant}}.to_json) }
+        let!(:unknown_identity) { Base64.encode64({'identity' => { 'account_number' => unknown_tenant}}.to_json) }
+      end
+    end
+  end
+end
```
Fryguy: missing CR at end of file.

app/controllers/application_controller.rb
gtanzillo: Yes, but I didn't think it would hurt to let it do that

spec/support/tenant_identity.rb
```diff
@@ -0,0 +1,15 @@
...
+  module Support
+    module TenantIdentity
+      extend ActiveSupport::Concern
+
+      included do
+        let!(:tenant)           { Tenant.create!(:external_tenant => external_tenant) }
+        let!(:external_tenant)  { rand(1000).to_s }
+        let!(:unknown_tenant)   { rand(1000).to_s }
+        let!(:identity)         { Base64.encode64({'identity' => { 'account_number' => external_tenant}}.to_json) }
+        let!(:unknown_identity) { Base64.encode64({'identity' => { 'account_number' => unknown_tenant}}.to_json) }
```
Fryguy: Feels like we should have helper methods in ManageIQ::API::Common for this...future follow up. cc @syncrou

Fryguy:
Looking good...got a number of comments above.

spec/support/tenant_identity.rb
syncrou: @Fryguy, @gtanzillo  - Yes.  We actually already do.  https://github.com/ManageIQ/manageiq-api-common/blob/master/spec/support/user_header_spec_helper.rb

syncrou: There are a number of examples in the manageiq-common and the catalog where a new key/value is applied and re Base64 encoded as well.

app/controllers/application_controller.rb
gtanzillo: I think I can get this to work. Looking at it now

spec/controllers/api/v0x0/application_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
+RSpec.describe ApplicationController, :type => :request do
```
agrare: `spec/controllers/api/v0x0/application_controller_spec.rb`

@gtanzillo should this be in `spec/controllers/application_controller_spec.rb` instead of under `api/v0x0` ?

gtanzillo: Yes, I'll move it

app/controllers/api/v0/sources_controller.rb
```diff
@@ -6,6 +6,17 @@ class SourcesController < ApplicationController
       include Api::V0::Mixins::ShowMixin
       include Api::V0::Mixins::UpdateMixin
 
+      def with_current_request
+        ManageIQ::API::Common::Request.with_request(request) do |current|
+          if Tenant.tenancy_enabled?
+            # Auto-create tenant if creating a new source and tenant does not exist
+            Tenant.find_or_create_by(:external_tenant => current.user.tenant) if action_name == "create" rescue KeyError
```
agrare: Do you need to set the current tenant here after creating it?

gtanzillo: Should be handled by `super`. But, in looking at that, I saw the the test was not quite correct.

PR #167, opened 2019-03-28
Add Application and ApplictionType controllers

agrare:
Add applications and application_types controllers

The modeling was originally added to core in https://github.com/ManageIQ/topological_inventory-core/pull/116

### DIFF
--- a/app/controllers/api/v0x1/application_types_controller.rb
+++ b/app/controllers/api/v0x1/application_types_controller.rb
@@ -1,0 +1,8 @@
+module Api
+  module V0x1
+    class ApplicationTypesController < ApplicationController
+      include Api::V0x1::Mixins::IndexMixin
+      include Api::V0::Mixins::ShowMixin
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


agrare:
cc @gtanzillo @abellotti

martinpovolny:
Oh, well. You could have pinged me. I actually started working on this one.

martinpovolny:
But never mind. From the learning perspective, I can read ;-)

A question: why aren't you also creating subcollections for the other relations such as application_type/application? Is it somehow systematic? Or is it what you think will be needed?

agrare:
> Oh, well. You could have pinged me. I actually started working on this one.

Oh sorry didn't know that, I was trying to unblock you haha

> A question: why aren't you also creating subcollections for the other relations such as application_type/application? Is it somehow systematic? Or is it what you think will be needed?

Do you mean why isn't there an application_types subcollection under the application route?  It is because application_type is a `belongs_to` on an `application` so you just get back a single `application_type_id`

Or do you mean why isn't there a `sources/application_types` using the `:has_many :through` relation?

martinpovolny:
>  application_types subcollection under the application route?

I mean the other way round

agrare:
These were just based on what I thought would be used, if you need to go from application_types to applications then we can add a route for it.

PR #170, opened 2019-03-29
Temporary fix for the starts_with? issue, where swagger patterns have already been converted to Regexp.

abellotti:


Fryguy:
Is it possible to create a spec that demonstrates the error?

agrare:
@Fryguy yeah I'm trying to get some reproduction steps...no luck so far

martinpovolny:
@agrare, @Fryguy :

I have a 100% way to reproduce. Actually I understand the error and I also happen to have a sort-of fix.

The problem is that a (Hash) structure is being updated from multiple threads in parallel, values are passed by reference. State mutation. Well Ruby allows you to do these things. Anyone willing to pay me to rewrite the API in Elixir? ;-)

Here's my fix: https://github.com/ManageIQ/topological_inventory-api/pull/171

However I am very unsure whether the logic of the `lib/open_api` should be done this way. I'd rather be more careful and do less state mutation in multi-threaded app.

martinpovolny:
To create a reproducer one needs to have the API running in a multi-threaded environment and fire 2 parallel requests that will cause the OpenAPI definitions to be parsed.

Good luck trying to do that in a unit test :-D

Well actually if you do that somehow, please, let me know, I'd love to learn some new dirty tricks ;-)

martinpovolny:
To elaborate a bit more on what I mean by "I'd rather be more careful":

I think that the algorithm in `lib/open_api` should be changed in such a way that instead of having one hash that is being changed in place, there would be two. One for input only (unprocessed parts of the schema) and the other one for the output (processed parts of the schema). The output values would be written once completely calculated, no intermediate values.

It that way even w/o locking and even if multiple threads would be parsing the same part of the schema, the result would be correct.

PR #171, opened 2019-04-01
Fix a state mutation race condition in substitute_references.

martinpovolny:
Because the `substitute_references` is called in parallel even the line
```
  object.each do |k, v|
    object[k] = ...
```
can get called in parallel. This results in the 2 (or more) threads
updating the structure passed in all the way by reference interfering
with each other.

For reference, the original problem's stacktrace:
```
[----] F, [2019-04-01T09:58:16.908016 #27866:1ab227c] FATAL -- : NoMethodError (undefined method `starts_with?' for /^\d+$/:Regexp):
[----] F, [2019-04-01T09:58:16.908057 #27866:1ab227c] FATAL -- :                                                                                                                                                  
[----] F, [2019-04-01T09:58:16.908102 #27866:1ab227c] FATAL -- : lib/open_api/docs/component_collection.rb:57:in `regexp_from_pattern'                                                                            
lib/open_api/docs/component_collection.rb:48:in `block in substitute_regexes'                                                                                                                                     
lib/open_api/docs/component_collection.rb:47:in `each'                                                                                                                                                            
lib/open_api/docs/component_collection.rb:47:in `substitute_regexes'                                                                                                                                              
lib/open_api/docs/component_collection.rb:19:in `load_definition'                                                                                                                                                 
lib/open_api/docs/component_collection.rb:12:in `[]'                                                                                                                                                              
lib/open_api/docs/component_collection.rb:39:in `public_send'                                                                                                                                                      lib/open_api/docs/component_collection.rb:39:in `fetch_ref_value'                                                                                                                    
lib/open_api/docs/component_collection.rb:30:in `substitute_references'                                                                                                                                           
lib/open_api/docs/component_collection.rb:31:in `block in substitute_references'                                                                                                                                   
lib/open_api/docs/component_collection.rb:31:in `each'                                                                                                                                                             
lib/open_api/docs/component_collection.rb:31:in `substitute_references'                                                                                                                                           
lib/open_api/docs/component_collection.rb:31:in `block in substitute_references'                                             
lib/open_api/docs/component_collection.rb:31:in `each'                             
lib/open_api/docs/component_collection.rb:31:in `substitute_references'                                                                                                                                           
lib/open_api/docs/component_collection.rb:20:in `load_definition'                                                                                                                                                 
lib/open_api/docs/component_collection.rb:12:in `[]'                                                                                                                                                              
app/controllers/application_controller.rb:30:in `api_doc_definition'                                                                                                                                              
app/controllers/application_controller.rb:152:in `api_doc_definition'                                                                                                                                              app/controllers/application_controller.rb:64:in `permitted_params'                                                                            
app/controllers/application_controller.rb:60:in `safe_params_for_list'                                                                                                                                            
app/controllers/application_controller.rb:124:in `filtered'                                                                                                                          
app/controllers/api/v0x1/mixins/index_mixin.rb:8:in `index'      
```

### DIFF
--- a/lib/open_api/docs/component_collection.rb
+++ b/lib/open_api/docs/component_collection.rb
@@ -43,8 +43,8 @@ module OpenApi
         if object.kind_of?(Array)
           object.collect { |i| substitute_regexes(i) }
         elsif object.kind_of?(Hash)
-          object.each do |k, v|
-            object[k] = k == "pattern" ? regexp_from_pattern(v) : substitute_regexes(v)
+          object.each_with_object({}) do |(k, v), o|
+            o[k] = k == "pattern" ? regexp_from_pattern(v) : substitute_regexes(v)
           end
         else
           object
@@ -52,7 +52,6 @@ module OpenApi
       end
 
       def regexp_from_pattern(pattern)
-        return pattern if pattern.kind_of?(Regexp)
         raise "Pattern #{pattern.inspect} is not a regular expression" unless pattern.starts_with?("/") && pattern.ends_with?("/")
         Regexp.new(pattern[1..-2])
       end


agrare:
Ah interesting, so both threads are trying to build a cached version on first access?  Could we just build the Api::Doc once on startup and treat it as readonly from then on?

martinpovolny:
> Ah interesting, so both threads are trying to build a cached version on first access? Could we just build the Api::Doc once on startup and treat it as readonly from then on?

You mean when starting the application?

I assume that (@bdunne ?) did the calculation on first use to speed up the start. But I guess it can be done. 

But regardless of future work I think that this PR should be merged. I think we should not be modifying a hash while iterating it.

agrare:
> You mean when starting the application?

> I assume that (@bdunne ?) did the calculation on first use to speed up the start. But I guess we can give that up.

I don't think it was to reduce startup time, I think it just wasn't considered that multiple threads would be trying to build that structure.

I'm good with this, do you also want to work on moving building this to an initializer?

martinpovolny:
> I'm good with this, do you also want to work on moving building this to an initializer?

Looking at how much work that is...

It seems that each controller for each API version is lazy-loading his part of the definition in the `API::Docs[version].definitions[name]`

Controller:
```
  @api_doc_definition ||= Api::Docs[api_version[1..-1].sub(/x/, ".")].definitions[model.name]
```

and each level of that is lazy loaded too:

DocV2:
```
  definitions ||= OpenApi::Docs::ComponentCollection.new(self, "definitions") 
```

ComponentCollection:
```
  def [](name)
    super || load_definition(name)
```

So on "moving this to the initializers", how would you preload that? 

Iterate over all the controllers and touch the structure?

Well I can start by changing ComponentCollection to preload all the definitions.

Looking at the code it's centered about the idea of lazy loading.

More fun: the `ComponentCollection` is a subclass of `Hash` which, as I remember, is a big NO-NO for a number of reasons. It's past my understanding why we are doing such things in new code.

Or do you have a better idea on how to approach this?

agrare:
> Iterate over all the controllers and touch the structure?
> 
> Well I can start by changing ComponentCollection to preload all the definitions.
> 
> Looking at the code it's centered about the idea of lazy loading.

Yeah I was thinking instead of lazy loading components we would iterate through an eager load every section.

PR #172, opened 2019-04-01
Add a tenants controller to the internal API

agrare:
Add a method for listing tenants on the internal API

Depends: https://github.com/ManageIQ/topological_inventory-core/pull/128

### DIFF
--- a/app/controllers/internal/v0/tenants_controller.rb
+++ b/app/controllers/internal/v0/tenants_controller.rb
@@ -1,0 +1,13 @@
+module Internal
+  module V0
+    class TenantsController < ::ApplicationController
+      def index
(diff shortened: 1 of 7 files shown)


agrare:
cc @abellotti @Ladas @gtanzillo

abellotti:
This LGTM @agrare. 👍 

On a related note, Since we're dropping all swagger 2.0/yaml's, for the internal v0.0 calls, I'll need to convert the v0.0.1 yaml to an openapi 3.0.0 json in https://github.com/ManageIQ/topological_inventory-api/pull/157 so we still support the /internal/v0.0 even though /api/v0.0 will be going away in that pr.

PR #175, opened 2019-04-03
Auto-create tenant on first request if it doesn't already exist

gtanzillo:


agrare:
```
       expected {"external_tenant" => "external_tenant_uuid", "id" => "146", "name" => "default"} to include {"id" => 146}
       Diff:
       @@ -1,4 +1,4 @@
        "external_tenant" => "external_tenant_uuid",
       -"id" => 146,
       +"id" => "146",
```

@gtanzillo that looks like a valid failure, should have been `"146" originally not sure why it wasn't

app/controllers/application_controller.rb
```diff
@@ -28,7 +28,7 @@ def with_current_request
     ManageIQ::API::Common::Request.with_request(request) do |current|
       begin
         if Tenant.tenancy_enabled?
-          tenant = Tenant.find_by(:external_tenant => current.user.tenant)
+          tenant = Tenant.find_or_create_by(:external_tenant => current.user.tenant) if current.user.tenant
```
Fryguy: I believe find_or_create_by is a multi-threaded race condition...need to deal with that or we will have duplicate tenants.

agrare: There's a unique constraint on tenants.external_tenant

PR #179, opened 2019-04-05
[WIP] Use the common routing helper for major version redirects

carbonin:
Requires https://github.com/ManageIQ/manageiq-api-common/pull/50

This also fixes the issue where a request to /openapi.json would
be redirected to /openapi


carbonin:
oops forgot #178 existed

PR #180, opened 2019-04-07
Enhanced the Tenant checks to know about optional auth requests

abellotti:
Enhanced the Tenant checks to know about request paths that are declared as having optional authentication, so requests for those from the gateway may or may not have the x-rh-identity header forwarded.

Depends on: https://github.com/ManageIQ/manageiq-api-common/pull/51

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,10 +28,12 @@ class ApplicationController < ActionController::API
     ManageIQ::API::Common::Request.with_request(request) do |current|
       begin
         if Tenant.tenancy_enabled?
-          tenant = Tenant.find_or_create_by(:external_tenant => current.user.tenant) if current.user.tenant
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -28,10 +28,12 @@ def with_current_request
     ManageIQ::API::Common::Request.with_request(request) do |current|
       begin
         if Tenant.tenancy_enabled?
```
gtanzillo: Could we just change this line to -

`if Tenant.tenancy_enabled? && current.required_auth?`

Then I don't think we'd need any other changes. It think...

abellotti: I think so and would  be fine for now (openapi.json use case), we can update later if we need to have paths that behave differently (with or without auth as the x-rh-identity could be there in case folks do authenticate).

abellotti:
there you go @gtanzillo, keepin' it simple.

PR #184, opened 2019-04-17
Defining and Enabling /api/v1.0 of the API

abellotti:
Enabling /api/v1.0
- OpenAPI 3.0.0  for v1.0
- Controllers (based on v0.1)
- Specs added for /api/v1.0

### DIFF
--- a/app/controllers/api/v1x0.rb
+++ b/app/controllers/api/v1x0.rb
@@ -1,0 +1,36 @@
+module Api
+  module V1x0
+    class RootController < ApplicationController
+      def openapi
+        render :json => Api::Docs["1.0"].content.to_json
+      end
+    end
+
+    class ApplicationsController                < Api::V0x1::ApplicationsController; end
+    class ApplicationTypesController            < Api::V0x1::ApplicationTypesController; end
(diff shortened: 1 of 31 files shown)


app/controllers/api/v1x0.rb
```diff
@@ -0,0 +1,97 @@
+module Api
+  module V1x0
+    class RootController < ApplicationController
+      def openapi
+        render :json => Api::Docs["1.0"].content.to_json
+      end
+    end
+
+    class AuthenticationsController < Api::V0x0::AuthenticationsController
+      include Api::V0x1::Mixins::IndexMixin
```
agrare: @abellotti could we also just drop `v0.0` and move the `v0x1` index_mixin into the base mixins?

abellotti: no one is using /api/v0.0  (just the /internal/v0.0), should I drop that in this PR ?

abellotti: ok, /api/v0.0 goes bye bye

abellotti:
ping @gtanzillo will take some work to remove /api/v0.0, let's do this in a followup-PR.

PR #185, opened 2019-04-17
Fix tests for tenancy enforcement

gtanzillo:


app/controllers/application_controller.rb
```diff
@@ -35,7 +35,7 @@ def with_current_request
         else
           ActsAsTenant.without_tenant { yield }
         end
-      rescue TopologicalInventory::Api::NoTenantError, KeyError
+      rescue TopologicalInventory::Api::NoTenantError, KeyError, ManageIQ::API::Common::IdentityError
```
agrare: Do we need our own NoTenantError anymore?  It looks like `current.user.tenant` might blow up before we get our own keyerror.

spec/requests/api/v0.0/authentications_spec.rb
```diff
@@ -1,12 +1,14 @@
 RSpec.describe("v0.0 - Authentications") do
+  include ::Spec::Support::TenantIdentity
+
+  let(:headers)         { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
   let(:attributes)      { {"username" => "test_name", "password" => "Test Password", "tenant_id" => tenant.id.to_s, "resource_type" => "Tenant", "resource_id" => tenant.id.to_s} }
   let(:collection_path) { "/api/v0.0/authentications" }
-  let(:tenant)          { Tenant.find_or_create_by!(:name => "default", :external_tenant => "external_tenant_uuid")}
 
-  describe("/api/v0.0/authentications") do
+  pending("/api/v0.0/authentications") do
```
agrare: Interesting, what fails here?

agrare:
Nice @gtanzillo !  Fixing all those specs must have gotten repetitive

app/controllers/application_controller.rb
gtanzillo: I think you are right. That error was raised before we started auto creating the tenant. Let me verify that.

spec/requests/api/v0.0/authentications_spec.rb
gtanzillo: That was a local change that shouldn't have been pushed. I've reversed it

app/controllers/application_controller.rb
gtanzillo: @agrare I pushed another commit to remove `NoTenantError`

agrare: :+1:

gtanzillo:
Thanks! @agrare

PR #187, opened 2019-04-25
Multilng client generator

slemrmartin:
rake task for generating client in more languages, defaulting ruby

### DIFF
--- a/lib/tasks/client_generate.rake
+++ b/lib/tasks/client_generate.rake
@@ -52,7 +52,7 @@ class ClientGenerator
     File.write(yaml_spec, JSON.parse(File.read(json_spec)).to_yaml(:line_width => -1).sub("---\n", "").tap { |c| c.gsub!("- NULL VALUE", "- null") })
   end
 
-  def generate_ruby_client(client_dir)
+  def generate_client(client_dir, lng)
     msg("Topological Inventory API Version: #{api_version}")
     msg("Using OpenAPI Generator CLI Jar:   #{generator_cli_jar}")
     msg("OpenAPI 3.0 Specification File:    #{openapi_file}")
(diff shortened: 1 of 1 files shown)


slemrmartin:
@abellotti please review

lib/tasks/client_generate.rake
```diff
@@ -61,15 +61,22 @@ class ClientGenerator
...
-  desc "Generate the Toplogical Inventory API Ruby Client"
+  desc "Generate the Topological Inventory API Ruby Client"
   task :generate, [:client_dir] => [:environment] do |_task, args|
     default_client_dir = Pathname.new(Rails.root.join("..", "topological_inventory-api-client-ruby"))
     args.with_defaults(:client_dir => default_client_dir)
-    ClientGenerator.new.generate_ruby_client(args[:client_dir])
+    ClientGenerator.new.generate_client(args[:client_dir], "ruby")
+  end
+
+  desc "Generate the Topological Inventory API Bash Client"
```
abellotti: I'd personally just use curl, but yea I think this is a great idea. Can you see if we can use the same rake task just add an optional lang argument defaulting to ruby.  This way we can simply use the same rake task for any other language without having to add a rake task for each ?  Thanks.

slemrmartin: @abellotti now it defaults to ruby, but second param can be added as language

abellotti: Nice update!! Thanks @slemrmartin 

Can you just update the usage in the comment section ? that'll great.
#        bundle exec rake client:generate[/alternate_ruby_client_dir]

Thanks.

slemrmartin: yes, and I'll create similar pr to sources-api too.

lib/tasks/client_generate.rake
```diff
@@ -61,15 +61,15 @@ class ClientGenerator
...
+    system("java -jar #{generator_cli_jar} generate -i #{openapi_yaml_file} -c #{generator_config} -g #{lng} -o #{client_dir}")
   end
 end
 
 namespace :client do
-  desc "Generate the Toplogical Inventory API Ruby Client"
-  task :generate, [:client_dir] => [:environment] do |_task, args|
+  desc "Generate the Topological Inventory API Ruby Client (default is Ruby)"
+  task :generate, [:client_dir, :language] => [:environment] do |_task, args|
     default_client_dir = Pathname.new(Rails.root.join("..", "topological_inventory-api-client-ruby"))
```
abellotti: hmm, this should probably change  to -api-client-#{language}

lib/tasks/client_generate.rake
slemrmartin: done

lib/tasks/client_generate.rake
slemrmartin: done

abellotti:
Thanks @slemrmartin for the update 👍 

/cc @agrare can you take a 👀

agrare:
Love the flexibility, is there anything we should add to the openapi config json slash do we need different config files per language?

slemrmartin:
@agrare I tried only bash client, it works correctly generated with the same json config

agrare:
Okay, something we should keep in mind if we go to generate e.g. python/go clients...they might need different json config.

PR #188, opened 2019-04-26
Return the proper error message for invalid URLs

abellotti:
Adding a catch-all in the routes so we can return the proper error message with the 404.


abellotti:
now returning the standard error object we return, and complying with IPP-12.

```
{
  "errors": [
    {
      "status": 404,
      "detail": "Invalid URL /bogus/url specified."
    }
  ]
}
```

PR #190, opened 2019-05-02
[WIP] Provide better filtering syntax check to the clients.

abellotti:
Provide better filtering syntax check to the clients.

Benefits of implementation complexity is TBD.


abellotti:
Marking as WIP

- [ ] finer granularity of filter syntax per collection
- [ ] provide available attribute keys 
- [ ] differentiate integer vs time vs string operators

abellotti:
fyi @mkanoor here's my playground.

PR #191, opened 2019-05-06
Sources API should only be exposed read-only.

abellotti:
Sources API should only be exposed read-only.
Routes and OpenAPI json spec was still allowing creates, updates and deletes even though the code did not.


agrare:
Thanks @abellotti!

PR #192, opened 2019-05-07
Remove Api::Docs.content.to_json

agrare:
The .content shouldn't be needed now that OpenApi::Docs to_json method
returns .content


agrare:
cc @Fryguy

abellotti:
👍 with the https://github.com/ManageIQ/manageiq-api-common/pull/65 counterpart checked in.  Thanks @agrare for this update.

PR #193, opened 2019-05-13
Fix for POST create openapi JSON signatures.

abellotti:
Fix for POST create openapi JSON signatures.  

Response signatures like:

         "content": {
           "application/json": {
             "schema": {
               "type": "object",
               "items": {
                 "$ref": "#/components/schemas/<<Collection>>"
               }
             }
           }
         }

  Should just be for simple json objects and not arrays:

         "content": {
           "application/json": {
             "schema": {
               "$ref": "#/components/schemas/<<Collection>>"
             }
           }
         }

Updating the openapi:generate tasks only as none of the entries in the topological inventory openapi spec included such endpoints, those were moved to the sources api.

### DIFF
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -297,10 +297,7 @@ class OpenapiGenerator
           "description" => "#{klass_name} creation successful",
           "content"     => {
             "application/json" => {
-              "schema" => {
-                "type"  => "object",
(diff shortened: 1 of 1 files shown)


Fryguy:
Is the generator not in common code?

abellotti:
not generic enough to be in common code, model class checks and specific schema, one-offs (icon data), etc.  differentiating topo from sources, etc.  Could be generalized I presume where callers can invoke with their customizations.

PR #194, opened 2019-05-18
Support GraphQL

abellotti:
* Things to do:
  * [x] Use templates for GraphQL schema definitions
  * [x] Support Associations
  * [x] Query by Attr/Filter
  * [x] Paging
  * [x] Honors x-rh-identity tenancy.
  * [x] Generalize code in ManageIQ::API::Common::GraphQL for usage by other services

### Adding Support for GraphQL

- Dynamic GraphQL Schema Generation
- Wanted the truth to be the OpenAPI specification as it is based on the model already
- OpenAPI specification also includes some information like attribute description that is not available in the model.
- No GraphiQL Integration here as that is more of a UI thing - Keeping this strictly GraphQL API Core.
- Simply use the graphiql client to try out:
  - brew cask install graphiql
  - run /Applications/GraphiQL.app/Contents/MacOS/GraphiQL
- Exposes /api/v1.0/graphql 

#### Usage:

\$ bundle exec rails s

Bring up GraphiQL and point to http://localhost:3000/api/v1.0/graphql

You can simply introspect the primary collections and related attributes and perform the query.


abellotti:
It's a beautiful thing ...

Screen Shot 2019-05-18 at 12 02 09 AM

abellotti:
Now with associations we can do queries like:

```
{
  container_groups {
    id
    name
    created_at
    containers {
      edges {
        node {
          id
          name
          archived_at
        }
      }
    }
  }
}
```

abellotti:
Paging implemented via Relay spec's connection model as per https://graphql.org/learn/pagination/ and https://facebook.github.io/relay/graphql/connections.htm

paging parameters including: first, last, after, before
pageInfo parameters including: startCursor, endCursor, hasNextPage, hasPreviousPage

pluralized model names (e.g. *:container_groups*) expose the Relay cursor connection, singularized expose a single resource fetch (e.g. *:container_group*).

Starting the graph with a single container group:

```
{
  container_group(id: 144) {
    id
    name
    source_ref
    containers(first: 5) {
      edges {
        node {
          id
          name
          archived_at
        }
      }
    }
  }
}
```

Paging through all container groups:

```
{
  container_groups(first: 20) {
    edges {
      node {
        id
        name
        source_ref
        containers {
          edges {
            node {
              id
              name
              archived_at
            }
          }
        }
      }
    }
    pageInfo {
      startCursor
      endCursor
    }
  }
}
```

abellotti:
**Added Support for Filtering**

The GraphQL language purposely excludes a filtering specification and leaves it up to the implementor to specify theirs based on the back-end implementations. While there are some third-party implementations like https://www.graphql.cool which supports flexible filtering, it didn't make sense for us to introduce yet another filtering syntax. It was best to simply expose the filter parameter to support our ManageIQ::Api::Common::Filter implementation used by our core services as per https://github.com/ManageIQ/manageiq-api-common#manageiqapicommonfilter

Usage simply extends the GraphQL relay connections with the filter parameter as follows:

```
{
  containers(filter: {id: {gt: "200"}}) {
    edges {
      node {
        id
        name
        container_group_id
      }
    }
  }
}
```

```
{
  container_groups(filter: {id: ["5", "10", "20"]}) {
    edges {
      node {
        id
        name
        containers (first: 100) {
          edges {
            node {
              id
              name
            }
          }
          pageInfo {
            hasNextPage
          }
        }
      }
    }
  }
}
```

abellotti:
**Filtering Examples:**

| Query / Ruby Client / GraphQL | 
|-------------------------------|
|"?filter[name]=reviews"                                | 
|`{:filter => { :name => "reviews" }}`                  |
|**`filter: { name: "reviews" }`** |
||
|"?filter[name][eq]=reviews"|
|`{:filter => { :name => { :eq => "reviews" } }}`|
|**`filter: { name: { eq: "reviews" } }`**|
||
|"?filter[name][starts_with]=a"|
|`{:filter => { :name => { :starts_with => "a" } }}`|
|**`filter: { name: { starts_with: "a" } }`**|
||
|"?filter[name][ends_with]=manager"|
|`{:filter => { :name => { :ends_with => "manager" } }}`|
|**`filter: { name: { ends_with: "manager" } }`**|
||
|"?filter[name][contains]=openshift"|
|`{:filter => { :name => { :contains => "openshift" } }}`|
|**`filter: { name: { contains: "openshift" } }`**|
||
|"?filter[id]=5"|
|`{:filter => { :id => "5" }}`|
|**`filter: { id: "5" }`**|
||
|"?filter[id][eq]=5"|
|`{:filter => { :id => { :eq => "5" }}}`|
|**`filter: { id: { eq: "5" } }`**|
||
|"?filter[id][gt]=180"|
|`{:filter => { :id => { :gt => "180" }}}`|
|**`filter: { id: { gt: "180" } }`**|
||
|"?filter[id][gte]=190"|
|`{:filter => { :id => { :gte => "190" }}}`|
|**`filter: { id: { gte: "190" } }`**|
||
|"?filter[id][lt]=5"|
|`{:filter => { :id => { :lt => "5" }}} `|
|**`filter: { id: { lt: "5" } }`**|
||
|"?filter[id][lte]=5"|
|`{:filter => { :id => { :lte => "5" }}}`|
|**`filter: { id: { lte: "5" } }`**|
||
|"?filter[id][]=5&filter[id][]=10&filter[id][]=15&filter[id][]=20"|
|`{:filter => { :id => ["5", "10", "15", "20"]}}`|
|**`filter: { id: ["5", "10", "15", "20"] }`**|
||
|"?filter[id][eq][]=5&filter[id][eq][]=10&filter[id][eq][]=15&filter[id][eq][]=20"|
|`{:filter => { :id => { :eq => ["5", "10", "15", "20"]}}}`|
|**`filter: { id: { eq: ["5", "10", "15", "20"] }`**|

abellotti:
Keeping as WIP as it depends on https://github.com/ManageIQ/manageiq-api-common/pull/69

public/doc/openapi-3-v1.0.0.json
```diff
@@ -806,6 +806,36 @@
         }
       }
     },
+    "/graphql": {
+      "post": {
+        "summary": "Perform a GraphQL Query",
+        "operationId": "postGraphQL",
+        "description": "Performs a GraphQL Query",
+        "requestBody": {
+          "content": {
```
eclarizio: @abellotti As we talked about, perhaps this section should include the format of the request as indicated by https://graphql.org/learn/serving-over-http/#post-request. I think you said this piece was auto-generated by the common gem, so it would likely have to change there instead of here.

abellotti: Thanks @eclarizio, I'll create follow-up PRs re: query/operationName/variables, as well as the data/errors in the response.

Also, according to  https://graphql.org/learn/serving-over-http/#post-request we could add support for the GET equivalent ie. GET /api/topological-inventory/v1.0/graphql?query={ ... }

abellotti:
Hi @Ladas, this implements the Relay spec, so cursor based pagination as per https://graphql.org/learn/pagination/

all that funky edges/node stuff, we need to sit chat with @Fryguy if that's what we want as signature, see third bullet in https://github.com/ManageIQ/manageiq-api-common/pull/69#issuecomment-497339404

anyways, for cursor based, it's essentially first: and after:, where you can also asked for cursor so you know where to query for next fetch:

```
{
  containers(first: 10) {
    edges {
      node {
        id
        name
      }
      cursor
    }
  }
}
```

that'll return something like:

```
{
  "data": {
    "containers": {
      "edges": [
        {
          "node": {
            "id": "144",
            "name": "recyler-container"
          },
          "cursor": "MQ"
        },
        {
          "node": {
            "id": "205",
            "name": "recyler-container"
          },
          "cursor": "Mg"
        },
        {
          "node": {
            "id": "7",
            "name": "details"
          },
          "cursor": "Mw"
        },
        ...
```

so if you want items after Mw above, it will be:

```
          "node": {
            "id": "7",
            "name": "details"
          },
          "cursor": "Mw"
        },
        ...
```

so if you want items after Mw above, it will be:

```
          "node": {
            "id": "7",
            "name": "details"
          },
          "cursor": "Mw"
        },
        ...
```

so if you want items after Mw above, it will be:

```
{
  containers(first: 10, after: "Mw") {
    edges {
      node {
        id
        name
      }
      cursor
    }
  }
}
```

You can also asked the general page info for that query via:

```
{
  containers(first: 10, after: "Mw") {
    edges {
      node {
        id
        name
      }
      cursor
    }
{
  containers(first: 10) {
    edges {
      node {
        id
        name
      }
      cursor
    }
    pageInfo {
      startCursor
      endCursor
      hasPreviousPage
      hasNextPage
    }
  }
}
```

Anyways, not sure if we're keeping this or we're rolling our own with offset/limit with the simpler signature.

Ladas:
@abellotti cool

there was an API conversation a while ago, where we agreed the cursor based approach is the most effective one and it should cover all our pagination usecases. So I don't see any reason for implementing limit/offset (that can be very heavy on io ops) when we already have the best one. :-) Also all APIs should support cursor/marker based pagination at some point

Ladas:
@abellotti also, we might need to paginate on the nested nodes too? E.g. if we'd have node with 10k tags. So I suppose we'd design it some standard way?

Also what about GraphQL and 3scale, do they play nice now?

abellotti:
connections also implement the relay specification, so the following is doable:

```
{
  container_groups(first: 2, after: "MTY") {
    edges {
      node {
        id
        name
        containers(first: 5, after: "MQ") {
          edges {
            node {
              id
              name
            }
            cursor
          }
        }
      }
      cursor
    }
    pageInfo {
      startCursor
      endCursor
      hasPreviousPage
      hasNextPage
    }
  }
}
```

w.r.t. 3scale, no official support for graphql that I can see. it is based on OpenAPI so we simply expose the /graphql endpoint in our OpenAPI spec. So like any other endpoint, we have the same protection via the Insights gateway w.r.t. Auth, and would be managed the same in 3scale. In the future, if/when we enable rate limiting in 3scale, there might be a concern since you can query multiple collections with GraphQL, would that be considered multiple queries in 3scale or the single one. 3scale would need to be enhanced to know about the introspected GraphQL schema and give us better management for it.  Again, may or may not be an issue for us.

Fryguy:
@abellotti Are we keeping the edges/nodes stuff?  I thought we discussed removing it, but I didn't see anything here about a need to keep it (perhaps other that it being Relay related)

abellotti:
We can keep it because of the cursor based pagination (and being well documented), but it is verbose for sure. Anyways, let's discuss in a meeting next week with @Ladas and @martinpovolny to look at pros/cons.

PR #196, opened 2019-06-06
Remove old unpaginated v0x0 API

agrare:
This removes the old original v0x0 API which was unpaginated.  This is in preparation for moving everything to the v1x0 API.


agrare:
cc @bdunne @abellotti @gtanzillo

PR #197, opened 2019-06-06
Move internal api v1x0

agrare:
We'll need to work with AIOPS to make sure they update any of their code which uses the internal API to get tenants

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -171,8 +171,13 @@ Rails.application.routes.draw do
 
   scope :as => :internal, :module => "internal", :path => "internal" do
     routing_helper.redirect_major_version("v0.0", "internal", :via => [:get])
+    routing_helper.redirect_major_version("v1.0", "internal", :via => [:get])
 
     namespace :v0x0, :path => "v0.0" do
+      resources :tenants, :only => [:index, :show]
+    end
+
+    namespace :v1x0, :path => "v1.0" do
       resources :tenants, :only => [:index, :show]
     end
   end
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -4074,9 +4074,7 @@
         "type": "object",
         "properties": {
           "id": {
-            "type": "string",
-            "readOnly": true,
-            "format": "uuid"
+            "$ref": "#/components/schemas/ID"
           },
           "name": {
             "type": "string",
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -4070,6 +4070,31 @@
           }
         }
       },
+      "Tenant": {
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": "string",
+            "readOnly": true,
+            "format": "uuid"
+          },
+          "name": {
+            "type": "string",
+            "readOnly": true,
+            "example": "Sample Tenant"
+          },
+          "description": {
+            "type": "string",
+            "readOnly": true,
+            "example": "Description of the Tenant"
+          },
+          "external_tenant": {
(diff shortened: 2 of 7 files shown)


agrare:
cc @Ladas @tumido are you guys using this endpoint?  We need to move it to paginated and v1.0

public/doc/openapi-3-v1.0.0.json
```diff
@@ -4005,6 +4005,32 @@
           }
         }
       },
+      "Tenant": {
```
bdunne: Hmm... I don't see any references to `Tenant` in the rest of the doc.  I wonder if we should have a separate doc or just allow all attributes for the internal API

config/routes.rb
```diff
@@ -164,9 +164,9 @@ def put(*_args); end
   end
 
   scope :as => :internal, :module => "internal", :path => "internal" do
-    routing_helper.redirect_major_version("v0.0", "internal", :via => [:get])
+    routing_helper.redirect_major_version("v1.0", "internal", :via => [:get])
```
bdunne: Should we also add a redirect for `/v0.0` to `/v1.0`?

bdunne: It's the same content that it used to be and I think it should be at v1.0.  In the future, I don't expect it to match the versioning of the public API.

public/doc/openapi-3-v1.0.0.json
agrare: Maybe? The best thing to do would probably be to have a separate openapi spec for the internal API, this is just what we've been doing already in the v0.0 openapi spec

config/routes.rb
agrare: Actually the reply will be different because the v0.0 version didn't use pagination

public/doc/openapi-3-v1.0.0.json
bdunne: Do we even need the `internal` API on Topo now that authentications have moved to Sources?  `tenants` feels like a thing that can be public as long as a logged in user can only see their tenant.

agrare: There was/is a requirement for internal services (AiOps) to get a list of all tenants so that they can then get all vms.  This definitely isn't the most efficient way of doing this and they're looking into better ways like requesting a dump of entire tables/databases to storage but currently they require this, cc @Ladas are you guys still using this?

Ladas: Yes, this is being used for current usecases. It's not clear yet, what will be the new way.

tumido:
Hi all, I'm about to change AI-Ops topo collector to adjust to this. Quick question: We use this API to collect the `external_tenant` property of each entry. Can you plese confirm that there's no change in the scheme? So all we need to change is an API version number? :wink:

agrare:
@tumido correct the scheme will not change, but this will be paged like all the other endpoints where before it was not.

tumido:
@agrare I see. We'll add pagination then :) Thank you :+1:

agrare:
@bdunne I updated this to only add the `/internal/v1.0` API, then once AIOPS updates their code we can drop the old v0 API.  This way we dont' have to coordinate perfectly.

public/doc/openapi-3-v1.0.0.json
```diff
@@ -4067,6 +4067,31 @@
...
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": "string",
```
bdunne: Should this be `"$ref": "#/components/schemas/ID"` instead?

agrare: :+1: done

agrare:
@tumido let me know when you've switched to using v1.0 so I can delete the older API

PR #198, opened 2019-06-06
Move v0x1 API to v1x0

agrare:
The v0.1 API has been superceeded by the v1.0 API and all callers have moved over to v1.0

Depends on:
~~https://github.com/ManageIQ/topological_inventory-api/pull/196~~


agrare:
Didn't delete the v0.0 openapi file so that moving the internal API to v1.0 is no longer a dependency for this

config/routes.rb
```diff
@@ -89,81 +88,6 @@ def put(*_args); end
       resources :volumes,            :only => [:index, :show]
     end
 
-    namespace :v0x1, :path => "v0.1" do
```
bdunne: Maybe not for this version, but in the future (v1.0+) we should probably return a `HTTP 410 Gone`

PR #203, opened 2019-06-20
Adjust regex pattern based on manageiq-api-common#79

bdunne:
https://github.com/ManageIQ/manageiq-api-common/pull/79


carbonin:
This will need a regenerated client, right?

bdunne:
It doesn't look like this change changes anything in the client

PR #205, opened 2019-06-20
Add endpoint for CFME manifest

Fryguy:
@agrare @bdunne Please review.


config/routes.rb
```diff
@@ -173,5 +173,9 @@ def put(*_args); end
     end
   end
 
+  namespace :cfme do
```
bdunne: To be publicly accessible, this needs to be somewhere inside the `api` scope.

Fryguy: Gah you're right...forgot about that...

PR #206, opened 2019-06-21
Updated the openapi generator to use the new graphql description

abellotti:
- Updated the openapi generator to leverage the new graphql description where the response is declared separately in the schema.


abellotti:
WIP - depends on https://github.com/ManageIQ/manageiq-api-common/pull/83

abellotti:
@miq-bot assign @bdunne

PR #209, opened 2019-07-03
Update the cfme manifest

agrare:
https://github.com/RedHatCloudForms/cfme-cloud_services/pull/10 update
the manifest, also update it on topological-inventory


agrare:
cc @Fryguy

PR #213, opened 2019-07-15
Choosing service plan order target

slemrmartin:
Based on Sources API source type.

Kafka queue's name suffix is chosen by `SourceType.name` (ServicePlan -> Source -> SourceType)

---

**related to** https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/12

### DIFF
--- a/app/controllers/api/v1/service_plans_controller.rb
+++ b/app/controllers/api/v1/service_plans_controller.rb
@@ -1,3 +1,5 @@
+require "sources-api-client"
+
 module Api
   module V1
     class ServicePlansController < ApplicationController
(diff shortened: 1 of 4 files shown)


slemrmartin:
@agrare please review

cc @abellotti

app/controllers/api/v1/service_plans_controller.rb
```diff
@@ -6,21 +8,59 @@ class ServicePlansController < ApplicationController
...
+
+      def retrieve_source_type(service_plan)
+        source = sources_api_client.show_source(service_plan.source_id.to_s)
+        if source.present?
+          source_type = sources_api_client.show_source_type(source.source_type_id.to_s)
+          if source_type.present?
+            #
+            # Process Source Type
+            #
+            yield source_type
```
agrare: I don't think the yield/block buys you anything besides increased indentation, just raise if the source_type is nil then return it

app/controllers/api/v1/service_plans_controller.rb
```diff
@@ -6,21 +8,59 @@ class ServicePlansController < ApplicationController
...
         head :bad_request
+      rescue StandardError => err
+        error_document = ManageIQ::API::Common::ErrorDocument.new.add(err.message)
+        render :json => error_document.to_h, :status => error_document.status
       end
 
       private
 
+      def sources_api_client
+        @sources_api_client ||=
```
agrare: I don't think you can cache this without including the x-rh-identity context, otherwise the tenant from the first API call will be used for every subsequent call.

PR #214, opened 2019-07-16
Add cluster.name and host.ems_ref to manifest

agrare:


config/cfme/manifest_5_11.json
```diff
@@ -142,6 +142,7 @@
       },
       "ems_clusters": {
         "id": null,
+	"name": null,
```
bdunne: The alignment looks to be off just a little bit

config/cfme/manifest_5_11.json
```diff
@@ -266,6 +271,7 @@
...
```
bdunne: same here

config/cfme/manifest_5_11.json
agrare: Hm you're right I think a tab made its way in there

agrare: Good catch, fixed

PR #217, opened 2019-07-17
Select the closest manifest to the version provided

agrare:
The CFME appliance will query for the manifest using the exact version
that it has (e.g. 5.11.0.0), we want to return the closest manifest to
that without having a manifest for every single version.

E.g.:
With the following manifests:
`[manifest_5_11.json, manifest_5_11_0.json, manifest_5_11_0_2]`

```
GET /cfme/manifest/5.11.0.0 => manifest_5_11_0.json
GET /cfme/manifest/5.11.0.2 => manifest_5_11_0_2.json
GET /cfme/manifest/5.11.1.0 => manifest_5_11.json
GET /cfme/manifest/5.11.2.0 => manifest_5_11.json
GET /cfme/manifest/5.10.4.0 => nil
```

### DIFF
--- a/lib/topological_inventory/api/cfme_manifest.rb
+++ b/lib/topological_inventory/api/cfme_manifest.rb
@@ -4,7 +4,11 @@ module TopologicalInventory
       def self.find(version)
         version.gsub!(".", "_")
 
-        manifest = nil
+        # Find the manifest file that best matches the version supplied.
+        # Look for a full match first then iteratively check for broader
+        # and broader versions.
+        #
+        # E.g. 5.11.0.0 will check for 5.11.0.0, then 5.11.0, 5.11, then 5
         until version.blank? || (manifest = manifest_by_version[version]).present?
           version = version.split("_")[0...-1].join("_")
         end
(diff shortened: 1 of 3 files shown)


agrare:
WIP pending tests

lib/topological_inventory/api/cfme_manifest.rb
```diff
@@ -0,0 +1,28 @@
...
+        until version.blank? || (manifest = manifest_by_version[version]).present?
+          version = version.split("_")[0...-1].join("_")
+        end
+
+        manifest
+      end
+
+      private_class_method def self.manifest_by_version
+        # List all manifests and cache the result to prevent a readdir on every API call
+        @manifest_by_version ||=
```
Fryguy: Just noticed this and the only downside is we can't add new versions without bouncing the app...even so, that's probably ok since we do a redeploy anyway.

agrare: Yeah I considered that but since they come from the filesystem on the container image I figured it would have to be rebuilt and bounced anyway.

If we want to pull them from a location which can be updated while running then we can change this.

PR #222, opened 2019-07-26
Fix ServerPlan.order kafka message

agrare:
After https://github.com/ManageIQ/topological_inventory-openshift/pull/92 changed the operations worker to use subscribe_topic not subscribe_messages ordering was broken because the received_message.message was nil while led to: `undefined method `split' for nil:NilClass`

The fix is to use publish_topic instead of publish_message.


syncrou:
LGTM 👍

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #224, opened 2019-07-30
Updated manifest for VMs

pemcg:
Update the manifest with a list of VM attributes need for the migration analytics project


config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
       "vms": {
         "id": null,
         "name": null,
+        "description": null,
         "type": null,
-        "guid": null,
+        "ems_ref": null,
```
agrare: This already exists [here](https://github.com/ManageIQ/topological_inventory-api/pull/224/files#diff-0633ae652eeb9f8a0956d9a21e33c2f9L34)

config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
...
         "uid_ems": null,
         "archived": null,
-        "cpu_cores_per_socket": null,
+        "orphaned": null,
```
agrare: I don't think that it makes sense to ask if a VM is orphaned when you're essentially doing `ems.vms` given that orphaned/archived means the ems_id is nil

config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
...
+        "retired": null,
```
agrare: Retired might make sense because we don't always delete retired VMs

agrare: Oh also it already exists [here](https://github.com/ManageIQ/topological_inventory-api/pull/224/files#diff-0633ae652eeb9f8a0956d9a21e33c2f9L43)

config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
...
         "cpu_total_cores": null,
+        "cpu_cores_per_socket": null,
```
agrare: This is already [here](https://github.com/ManageIQ/topological_inventory-api/pull/224/files#diff-0633ae652eeb9f8a0956d9a21e33c2f9L31)

config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
...
         "disks_aligned": null,
-        "ems_ref": null,
         "has_rdm_disk": null,
-        "host": {
-          "ems_ref": null
-        },
```
agrare: We need the `host.ems_ref` to connect VMs to hosts

config/cfme/manifest_5_11.json
```diff
@@ -24,58 +24,64 @@
...
-        "linked_clone": null,
-        "orphaned": null,
         "power_state": null,
+        "ipaddresses": null,
+        "hostnames": null,
```
agrare: I think these two are redundant if we're getting ipaddress and hostname from networks, these are essentially just `vm.hardware.networks.ipaddress` and `vm.hardware.networks.hostname`

config/cfme/manifest_5_11.json
pemcg: I removed this for the ManageIQ::Providers::OpenStack::CloudManager provider as they don't have hosts, but thinking about it they do have host aggregates which are sort of cluster-like. Should I add host aggregate for this provider type?

config/cfme/manifest_5_11.json
pemcg: Fair enough, I'm used to Vm.all where you do need to check these, I hadn't realised that the ems_id was nil in that case. I'll remove

config/cfme/manifest_5_11.json
agrare: Ah okay didn't see it was only removed for openstack, I'm good with this then.

PR #229, opened 2019-08-06
Fix in permitting correct task's context

slemrmartin:
When sending `Task.context` in correct format then params are parsed by controller as
`<ActionController::Parameters {"context"=><ActionController::Parameters {"service_instance"=>{"source_id"=>"4", "source_ref"=>162}, "remote_status"=>"running"} permitted: false>, "state"=>"running", "status"=>"ok"} permitted: false>`

Then `body_params.permit("context")` is ignored, common `params_for_update` doesn't work.

**depends on** https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/15
**depends on** https://github.com/ManageIQ/topological_inventory-openshift/pull/96
**depends on** https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/20

(cyclic dependencies)

### DIFF
--- a/app/controllers/api/v1/tasks_controller.rb
+++ b/app/controllers/api/v1/tasks_controller.rb
@@ -22,7 +22,7 @@ module Api
         permitted = api_doc_definition.all_attributes - api_doc_definition.read_only_attributes
         if body_params['context'].present?
           permitted.delete('context')
-          permitted << { "context" => {} }
+          permitted << {'context'=>{}}
         end
         body_params.permit(*permitted)
       end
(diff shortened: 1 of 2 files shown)


slemrmartin:
cc @syncrou

app/controllers/api/v1/tasks_controller.rb
```diff
@@ -15,6 +15,17 @@ def update
...
         head :no_content
       end
+
+      private
+
+      def params_for_update
+        permitted = api_doc_definition.all_attributes - api_doc_definition.read_only_attributes
+        if body_params['context'].present?
+          permitted.delete('context')
+          permitted << { "context" => {} }
```
bdunne: Do you mind changing to no spaces inside the braces for consistency?

slemrmartin: do you think `permitted << {'context'=>{}}` ? I would be ok with it, but rubocop isn't.

PR #235, opened 2019-08-19
ServiceOffering ordering

slemrmartin:
Provides ordering of offerings with or without plan(form)

- old way
  - `POST /service_plans/:service_plan_id/order "{ \"service_parameters\": {...}", \"provider_controler_parameters\":{}}"`
- new way
  - `POST /service_offerings/:service_offering_id/order "{\"provider_controler_parameters\":{}}`
  - `POST /service_offerings/:service_offering_id/order "{\"service_parameters\": {...}", \"provider_controler_parameters\":{}}"`

**depends on**:
- [x] https://github.com/ManageIQ/topological_inventory-openshift/pull/99
- [x] https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/24

**dependents**:
https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/16

PR #236, opened 2019-08-23
Added optional service_plan_id to ServiceOffering#order

slemrmartin:
ServicePlan's id is not important for ansible tower ordering, but it's needed for openshift operations, where it's ordered by SvcOffering.name and SvcPlan.name

Fixes https://github.com/ManageIQ/topological_inventory-api/pull/235

### DIFF
--- a/app/controllers/api/v1/service_offerings_controller.rb
+++ b/app/controllers/api/v1/service_offerings_controller.rb
@@ -30,13 +30,14 @@ module Api
       def params_for_order
         @params_for_order ||= params.permit(
           :service_offering_id,
+          :service_plan_id,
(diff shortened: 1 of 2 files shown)


slemrmartin:
@agrare pls :eyes: 
cc @Hyperkid123

slemrmartin:
@miq-bot add-label bug

agrare:
@slemrmartin Looks good, can you update the openapi json OrderParameters to include the service_plan_id?  Not sure if we need a different object for ServiceOfferingOrderParameters vs ServicePlanOrderParameters since it doesn't make sense to have a service_plan_id for a ServicePlan#order

PR #243, opened 2019-09-16
Remove redundant svc offering openapi params

slemrmartin:
`OrderParametersServiceOffering` doesn't need `service_offering_id` in properties because this is already defined as `id`.
The same for `OrderParametersServicePlan` and `service_plan_id`.
And `required: false` is default value so it's useless definition.

- **dependent**: https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/19

### DIFF
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -406,15 +406,34 @@ class OpenapiGenerator
       }
     }
 
-    schemas["OrderParameters"] = {
-      "type" => "object",
-      "properties" => {
-        "service_parameters" => {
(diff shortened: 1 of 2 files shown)


slemrmartin:
@agrare please :eyes:

agrare:
Hey @slemrmartin I forgot on your last PR but we need to update the openapi:generate rake task to match the new split order parameters https://github.com/ManageIQ/topological_inventory-api/blob/master/lib/tasks/openapi_generate.rake#L409-L421

slemrmartin:
@agrare I didn't realize it, thanks, it's there now

PR #250, opened 2019-09-26
Add service inventories

Ladas:


config/routes.rb
```diff
@@ -43,6 +43,9 @@ def put(*_args); end
       resources :datastores,              :only => [:index, :show]
       resources :flavors,                 :only => [:index, :show]
       resources :service_instances,       :only => [:index, :show]
+      resources :service_inventories,     :only => [:index, :show] do
+         resources :tags, :only => [:index]
```
agrare: Looks like rubocop is right here, there's 3 spaces

Ladas: :-o

Ladas: fixed

PR #252, opened 2019-10-01
Moved the GraphQL request in the OpenAPI doc to a separate schema entry.

abellotti:
related to: https://github.com/ManageIQ/manageiq-api-common/pull/110

### DIFF
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -962,27 +962,7 @@
           "content": {
             "application/json": {
               "schema": {
-                "type": "object",
-                "properties": {
-                  "query": {
-                    "type": "string",
-                    "description": "The GraphQL query",
-                    "default": "{}"
-                  },
-                  "operationName": {
-                    "type": "string",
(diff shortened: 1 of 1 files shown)


public/doc/openapi-3-v1.0.0.json
```diff
@@ -962,27 +962,7 @@
...
-                  "variables": {
-                    "type": "object",
-                    "description": "Optional Query variables",
-                    "nullable": true
-                  }
-                },
-                "required": [
-                  "query"
-                ]
+                "$ref": "#/components/schemas/GraphQLRequest"
```
Ladas: @abellotti actually generator seems to be deleting this line now

bdunne: @abellotti I think you need to update the `manageiq-api-common` dependency to 1.0.0 to pull in those changes, right?

abellotti: I missed this here somehow https://github.com/ManageIQ/manageiq-api-common/pull/120

PR #257, opened 2019-10-22
API endpoint for svc offering's applied inventories

slemrmartin:
Task TPINVTRY-701

---

- [x] **depends on**: https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/41

### DIFF
--- a/app/controllers/api/v1/service_offerings_controller.rb
+++ b/app/controllers/api/v1/service_offerings_controller.rb
@@ -10,8 +10,8 @@ module Api
       end
 
       # POST
-      def approval_inventories
-        send_message_and_generate_task_for(:approval_inventories)
+      def applied_inventories
+        send_message_and_generate_task_for(:applied_inventories)
       end
 
       private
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/service_offerings_controller.rb
```diff
@@ -36,6 +48,13 @@ def params_for_order
         ).to_h
       end
 
+      def params_for_applied_inventories
+        @params_for_applied_inventories ||= params.permit(
+          :service_offering_id,
+          :service_parameters => {}
```
agrare: Why do you need service_parameters?

slemrmartin: it should contain "Prompt on launch" inventory from catalog. As I posted comment with presentation to "Tag based Approvals" doc, computing of applied inventories without knowledge of user input should return wrong inventories list. 
So I rather prepared full version of operation now. It means that it doesn't fail if Approval doesn't provide it, but results won't be correct.

PR #259, opened 2019-10-31
Update tagging APIs

bdunne:
- Display the tag string
- Related to TPINVTRY-716

Depends on:
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/140
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/139
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/159
- [x] https://github.com/ManageIQ/insights-api-common-rails/pull/136
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/157
- [x] https://github.com/ManageIQ/manageiq-api-common/pull/131

Examples:
 `GET /api/v1.0/container_nodes/2/tags`
```
{
  "meta": {
    "count": 2,
    "limit": 100,
    "offset": 0
  },
  "links": {
    "first": "/api/v1.0/container_nodes/2/tags?offset=0",
    "last": "/api/v1.0/container_nodes/2/tags?offset=0"
  },
  "data": [
    {
      "tag": "/bdunne/key1=value1"
    },
    {
      "tag": "/bdunne/key_only_no_value"
    }
  ]
}
```

`POST /api/v2.0/things/{id}/tag` with body `[{"tag": "/a/b/c=1"}]`
Response `[{"tag": "/a/b/c=1"}]`

POST /api/v2.0/things/{id}/untag with body {"tag" => "/a/b/c=1"}


public/doc/openapi-3-v1.0.0.json
```diff
@@ -255,11 +255,11 @@
         ],
         "responses": {
           "200": {
-            "description": "Tags collection",
```
Fryguy: I think the external documentation should still say "tags".  taggings is an internal concept.

bdunne: I wanted people to know that this is different from the `/tags` endpoint

Fryguy: but it is the tags endpoint? `/container_groups/{id}/tags`

bdunne: But the `/tags` primary collection has a different response.  Do we want to change that in this PR as well?

Currently: `GET /api/v1.0/tags`
```
{
  "meta": {
    "count": 2,
    "limit": 100,
    "offset": 0
  },
  "links": {
    "first": "/api/v1.0/tags?offset=0",
    "last": "/api/v1.0/tags?offset=0"
  },
  "data": [
    {
      "created_at": "2019-10-29T19:37:17Z",
      "id": "1",
      "name": "key1",
      "namespace": "bdunne",
      "value": "value1"
    },
    {
      "created_at": "2019-10-29T19:37:26Z",
      "id": "2",
      "name": "key_only_no_value",
      "namespace": "bdunne",
      "value": ""
    }
  ]
}
```

bdunne: Done in this PR, created `/api/v2.0/...`

app/controllers/api/v2x0.rb
```diff
@@ -0,0 +1,82 @@
...
+        tag = Tag.find_by!(Tag.parse(body_params["tag"]))
+        primary_instance.tags.destroy(tag)
+
+        render :json => tag, :status => :no_content, :location => "#{instance_link(primary_instance)}/tags"
+      end
+
+      private
+
+      def model
+        primary_collection_model.tagging_relation_name.to_s.singularize.classify.safe_constantize
```
agrare: Easy for you to say :laughing:

app/controllers/api/v2x0.rb
```diff
@@ -0,0 +1,82 @@
...
+    class SubscriptionsController               < Api::V1x0::SubscriptionsController; end
+    class TagsController                        < Api::V1x0::TagsController; end
+    class TasksController                       < Api::V1x0::TasksController; end
+    class VmsController                         < Api::V1x0::VmsController; end
+    class VolumeAttachmentsController           < Api::V1x0::VolumeAttachmentsController; end
+    class VolumesController                     < Api::V1x0::VolumesController; end
+    class VolumeTypesController                 < Api::V1x0::VolumeTypesController; end
+
+    # New / changed in api/v2.0/
+    class TaggingsController < ApplicationController
```
agrare: Any reason this is defined in the base `v2x0` module and not e.g. `app/controllers/api/v2/taggings_controller.rb` ?

app/controllers/api/v2x0.rb
```diff
@@ -0,0 +1,82 @@
+module Api
+  module V2x0
+    class RootController < ApplicationController
+      def openapi
+        render :json => ::Insights::API::Common::OpenApi::Docs.instance["2.0"].to_json
+      end
+    end
+
+    # Same behavior as api/v1.0/
+    class AvailabilitiesController              < Api::V1x0::AvailabilitiesController; end
```
agrare: Do we want to inherit this from `Api::V1::Avail...` instead of `V1x0` ?  Theoretically if we released a `V1x1` we'd want the most recent in v2 right?

agrare:
@abellotti can you review?

app/controllers/api/v2x0.rb
bdunne: I moved it to `app/controllers/api/v2x0/taggings_controller.rb`.  I don't see the point of having a `v2` and a `v2x0`, `v2x0` can be the new base.  Any changes in there should be inherited in future versions.

app/controllers/api/v2x0.rb
bdunne: `V1x0` inherits from `V1` and I think `V2x0` should inherit from the latest version 1 controllers which are currently `V1x0`.  In the future if we release a version 1.1, we should update the `V2x0` controllers to inherit from that.

app/controllers/api/v2x0.rb
agrare: :+1:

app/controllers/api/v2x0.rb
agrare: Would `Api::V1` _always_ be the latest in the `1.*` line though?  I'm thinking we wouldn't have to update anything in `V2x0` if we updated to a `V1x1`.

This is mostly a moot point since I doubt we'll ever update v1 if we're going to deprecate it, mostly trying to figure out the process going forward

bdunne: It's currently the other way around, but it would be nice to not have to update it.
```
module Api
  module V1x0
    class AvailabilitiesController              < Api::V1::AvailabilitiesController; end
    class ClustersController                    < Api::V1::ClustersController; end
    class ContainerGroupsController             < Api::V1::ContainerGroupsController; end
```
> This is mostly a moot point since I doubt we'll ever update v1 if we're going to deprecate it, mostly trying to figure out the process going forward

I agree, always good to review the process and make improvements along the way.

app/controllers/api/v2x0/taggings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      def self.presentation_name
+        "Tag".freeze
+      end
+
+      def create
+        primary_instance = primary_collection_model.find(request_path_parts["primary_collection_id"])
+        tag = Tag.find_or_create_by!(Tag.parse(params_for_create["tag"]))
+        primary_instance.tags << tag
```
lindgrenj6: Found when adding functionality to Catalog (https://github.com/RedHatInsights/catalog-api/pull/495):
This fails if the tag already exists on `primary_instance`, I just added a guard clause and it went away.

app/controllers/api/v2x0/taggings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        render :json => tag, :status => :created, :location => "#{instance_link(primary_instance)}/tags"
+      end
+
+      def destroy
+        primary_instance = primary_collection_model.find(request_path_parts["primary_collection_id"])
+        tag = Tag.find_by!(Tag.parse(body_params["tag"]))
```
lindgrenj6: Found when adding functionality to Catalog (https://github.com/RedHatInsights/catalog-api/pull/495):
OpenAPI is not ok with having `DELETE` operations that use request bodies, it's not part of the HTTP spec. It also can't be tested with rspec, per here: https://github.com/zipmark/rspec_api_documentation/issues/368

Not sure where we want to go with this one, maybe just allow deleting on the subresource?

abellotti: well, HTTP spec does not prohibit it.  There seems to be a PR for the OpenAPI 3 https://github.com/OAI/OpenAPI-Specification/pull/1937 to add support for it.

app/controllers/api/v2x0/taggings_controller.rb
bdunne: I wonder if we should return a `409 Conflict` instead?

app/controllers/api/v2x0/taggings_controller.rb
bdunne: > It also can't be tested with rspec

I was able to add tests [here](https://github.com/RedHatInsights/topological_inventory-api/pull/259/commits/e1f3321dbbd3282e43c35335d9e4e9f363bf125e#diff-af8561cfcf1754b19f858109d77eca64R45)

bdunne: > Not sure where we want to go with this one, maybe just allow deleting on the subresource?

I agree that it is unconventional to DELETE with a request body.  Unfortunately the user doesn't know the id of the tag join record that we're displaying.  (and I don't think they should see it either)

lindgrenj6: @bdunne looking at your specs showed me where I was wrong - when you pass `:params` a hash in rspec on a POST it implicitly must call `.to_json` on it, when I specify that explicitly for my DELETE requests my specs work again. That makes me feel so much better, I'm ok with it. 

Hopefully that change gets merged so we can have the requestBody defined in the OpenAPI schema for the DELETE operation.

app/controllers/api/v2x0/taggings_controller.rb
bdunne: @lindgrenj6 I wound up returning a 304 here if it is already tagged

lindgrenj6: :+1: sounds good to me! I'll get that added to my implementation over on Catalog.

mkanoor:
@bdunne We had a discussion on the "DELETE" verb using the request body with @Fryguy . Since its atypical to have a DELETE contain a body it breaks the openapi validation and will run into issues with generated clients. When you run the validation on the openapi spec file attached with this PR we get the following errors
```
openapi-generator validate -i openapi-3-v2.0.0.json 
Validating spec (openapi-3-v2.0.0.json)
Errors:
	-attribute paths.'/container_groups/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/container_nodes/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/service_inventories/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/container_images/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/container_templates/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/network_adapters/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/subnets/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/security_groups/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/ipaddresses/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/networks/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/container_projects/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/service_offerings/{id}/tags'(delete).request_body is unexpected
	-attribute paths.'/vms/{id}/tags'(delete).request_body is unexpected

[error] Spec has 13 errors.
```

@Fryguy mentioned that we should be using the POST to the delete tags to resolve this issue.

bdunne:
> We had a discussion on the "DELETE" verb using the request body with @Fryguy

Is that a chat conversation that I can read somewhere? (I don't see anything in slack)

> Since its atypical to have a DELETE contain a body it breaks the openapi validation

That's unfortunate, it's [one of the recommendations in JSON API](https://jsonapi.org/format/#crud-updating-to-many-relationships).  Do you know if this will be fixed?



It feels strange to me to use POST to delete things.  To me POST implies create or take an action on a thing.
> By design, the POST request method requests that a web server accepts the data enclosed in the body of the request message, most likely for storing it. It is often used when uploading a file or when submitting a completed web form. 


Alternatively, I would be open to something like below, but both of these seem worse to me than DELETE with a request body:
`PATCH /api/v2.0/things/1/tags` with a body including all of the tags that should exist on that record
`DELETE /api/v2.0/things/1/tags?filter[tags][]=/a/b/c=1` (although that tag string will have to be properly escaped)

mkanoor:
@bdunne Jason had mentioned this In a meeting with Catalog and Approval teams. @gmcculloug @syncrou @Fryguy were present. We haven't seen any progress on the openapi spec for sending body with DELETE. https://github.com/OAI/OpenAPI-Specification/issues/1801

bdunne:
Updated based on the discussion yesterday...
`GET /api/v2.0/things/{id}/tags`
`POST /api/v2.0/things/{id}/tag` with body `{"tag" => "/a/b/c=1"}`
`POST /api/v2.0/things/{id}/untag` with body `{"tag" => "/a/b/c=1"}`

mkanoor:
@bdunne 
During 
POST /api/v2.0/things/{id}/tag with body {"tag" => "/a/b/c=1"}
If the tag doesn't exist, will the system create the tag first.

bdunne:
> @bdunne
> During
> POST /api/v2.0/things/{id}/tag with body {"tag" => "/a/b/c=1"}
> If the tag doesn't exist, will the system create the tag first.

Yes, it will

mkanoor:
@bdunne 

The "#/components/schemas/Tag" has a tag attribute which is marked as readonly and it is used in a requestBody in a post, should we remove the readonly?
``` 
"requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Tag"
              }
            }
          }
```

mkanoor:
@bdunne can you give an example of getting tags with a filter what would the URL look like?

GET /api/v2.0/things/{id}/tags?filter=tag[:tag => /ns/key=value]

abellotti:
how can we tag or untag multiple tags in a single POST ?  I see GET .../tags returns the multiple but what about .../tag and .../untag ?  should we support in the body signature tag : [ "tag1", "tag2", ... ]  or maybe the plural { "tags" : [ "tag1", "tag2", ... ] } ?

bdunne:
> how can we tag or untag multiple tags in a single POST ? I see GET .../tags returns the multiple but what about .../tag and .../untag ? should we support in the body signature tag : [ "tag1", "tag2", ... ] or maybe the plural { "tags" : [ "tag1", "tag2", ... ] } ?

Right now all of our POST actions are only accepting a single payload, so I was staying consistent with that.  I'd like to handle multiple tags in a separate PR, what do you think of that @abellotti ?

abellotti:
Hi @bdunne, looks good, the only thing that may be confusing is for the multi-tagging request, the response only returns the newly created one, one may expect the response to have them all specially if we return other attributes later with a tag that folks may be depending on in the response.

bdunne:
> Hi @bdunne, looks good, the only thing that may be confusing is for the multi-tagging request, the response only returns the newly created one, one may expect the response to have them all specially if we return other attributes later with a tag that folks may be depending on in the response.

Does "have them all" refer to all of the tags that were requested to be added or all of the tags that the thing is tagged with?  I thought returning the list of tags added would be more helpful in the response.

abellotti:
what I meant was, user requests tags a b and c added, I'd think the expectation is that he gets a b and c in the response.  the resource may be additional tags d and e, but we're not returning those, just the ones being requested.

bdunne:
@abellotti Done

PR #266, opened 2019-12-18
Enabling the sort_by parameter

abellotti:
Enabling the sort_by parameter, was only exposed via graphql for all, for rest, only the common dummy app in common had it exposed.  

Added a small spec to verify it is now exposed here, the comprehensive tests for sort_by is in common.

Fixes: TPINVTRY-801


abellotti:
Related PR for Sources: https://github.com/RedHatInsights/sources-api/pull/144

PR #268, opened 2020-01-08
Move openapi.json files to just x.y version numbers

agrare:
As we start updating the .z version of the API to push api-client gems
we shouldn't have to keep renaming the files for minor enhancements and
bugfixes.


agrare:
The plan is that every change to the openapi-3-v2.0.json file will rev the 2.0.z api version, then when we generate the api-client gems we will tag them with that version and push them to rubygems.

Only the latest .z version of the API will be accessible on the API server

agrare:
If we're good with this style, I'll change sources-api's openapi-3-v1.0.0.json and update the insights-api-common openapi::generator

PR #269, opened 2020-01-10
Internal API doesn't allow validation against openapi.json

slemrmartin:
OpenAPI.json doesn't contain internal API paths, so updating of `Source.refresh_status` returned 400 Bad Request error.


agrare:
Nice catch, I wonder if we should do this with all of the internal controllers?

PR #272, opened 2020-01-16
Enable Tag filtering

bdunne:
Depends on:
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/149
- [x] https://github.com/RedHatInsights/topological_inventory-core/pull/165


bdunne:
ok @abellotti @gtanzillo This should go green now.  I updated the dependency on the new insights-api-common and changed the v2.0 to v3.0.

PR #275, opened 2020-01-29
Bump puma to 3.12.2

abellotti:
- taking care of a Hakiri warning.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ gem 'manageiq-password',   '~> 0.2', ">= 0.2.1"
 gem 'mimemagic',           '~> 0.3.3'
 gem 'more_core_extensions'
 gem 'pg',                  '~> 1.0', :require => false
-gem 'puma',                '~> 3.0'
+gem 'puma',                '~> 3.12.2'
 gem 'rack-cors',           '>= 1.0.4', '~> 1.0'
 gem 'rails',               '>= 5.2.2.1', '~> 5.2.2'
 


Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'manageiq-password',   '~> 0.2', ">= 0.2.1"
 gem 'mimemagic',           '~> 0.3.3'
 gem 'more_core_extensions'
 gem 'pg',                  '~> 1.0', :require => false
-gem 'puma',                '~> 3.0'
```
bdunne: Do we need to lock down to 3.12?  Is `~>3.0, >=3.12.2` ok?

abellotti: I think that would work too.  was trying to be consistent with sources.

abellotti: Somehow Hakiri did not like ~>3.0, >=3.12.2,  moving to just >= 3.12.2 worked.

agrare: If I remember right hakiri wanted `>= 3.12.2` first, then `~> 3.0`

abellotti: Ahh, lemme try that.

agrare: Wow really can't make hakiri happy...

abellotti: holy cow, REALLY picky, tried '>= 3.12.2' with '~> 3.0', then '~> 3.12', and finally with '~> 3.12.2', still failed.  looks like the only one that works is the original one like sources, i.e. just '~> 3.12.2'

bdunne:
Any reason we shouldn't upgrade to v4? https://github.com/RedHatInsights/topological_inventory-api/pull/276

bdunne:
Closing in favor of #276

PR #278, opened 2020-02-04
Updated list of manifest attributes

pemcg:
Added 4 new attributes which will help identify which workloads are unsuitable for migration to CNV


pemcg:
@miq-bot add_label enhancement

PR #281, opened 2020-02-18
removed guest_applications and added disk size attributes

pemcg:
This PR removes the guest_applications section because it makes the payload size too large. It also adds disk storage size attributes so that allocated and used disk sized can be displayed in a Migration Analytics report.

The last_scan_on attribute is needed to determine whether a SmartState Analysis has been performed. If it hasn't then allocated_disk_storage will be the same as used_disk_storage, regardless of whether the disks are or aren't thin provisioned.

### DIFF
--- a/config/cfme/manifest_5_11.json
+++ b/config/cfme/manifest_5_11.json
@@ -103,8 +103,12 @@
         "description": null,
         "type": null,
         "uid_ems": null,
+        "last_scan_on": null,
         "cpu_cores_per_socket": null,
         "cpu_total_cores": null,
+        "allocated_disk_storage": null,
+        "used_disk_storage": null,
+        "thin_provisioned": null,
         "disks_aligned": null,
         "ems_ref": null,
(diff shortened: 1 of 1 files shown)


pemcg:
@miq-bot add_label enhancement

config/cfme/manifest_5_11.json
```diff
@@ -214,7 +211,6 @@
         "uncommitted": null,
         "storage_domain_type": null,
         "host_storages": {
-          "ems_ref": null,
```
agrare: Hey @pemcg so on 5.11 this is actually needed, there is no 5.12 manifest because the sync feature was removed
The column was removed on master but is still there on 5.12

gmcculloug:
@pemcg Are there any plans to move forward with this PR and address the issue Adam raised or should it be closed?

pemcg:
This can be closed, I think the changes were implemented in other PRs. Thanks Greg

PR #282, opened 2020-02-20
[WIP] Moved the SortBy parameter as an Object definition in the OpenAPI spec

abellotti:
Moved the SortBy parameter as an Object definition in the schema similar to the Filter.  This is to get around an issue with the OpenAPI client generator not properly dealing with the OneOf Schema syntax.

Depends on: https://github.com/RedHatInsights/insights-api-common-rails/pull/160


abellotti:
Closing in preference to: https://github.com/RedHatInsights/topological_inventory-api/pull/291

PR #284, opened 2020-02-24
Add credential and credential types endpoint

lpichler:
TPINVTRY-726
Api endpoints for service credential and service  credential types
and  credentials subcollections for service_offerings,service_instances,. ..

### DIFF
--- a/app/controllers/api/v1/mixins/index_mixin.rb
+++ b/app/controllers/api/v1/mixins/index_mixin.rb
@@ -12,20 +12,6 @@ module Api
             :sort_by    => query_sort_by
           ).response
         end
-
-        def scoped(relation)
-          if through_relation_klass
-            relation = relation.joins(through_relation_name)
-          end
-
-          relation
-        end
-
-        def raise_unless_primary_instance_exists
-          return unless subcollection?
-
-          primary_instance
-        end
       end
     end
   end
--- a/app/controllers/api/v3x0/service_credential_types_controller.rb
+++ b/app/controllers/api/v3x0/service_credential_types_controller.rb
@@ -1,0 +1,9 @@
+module Api
+  module V3x0
+    class ServiceCredentialTypesController < ApplicationController
+      include Mixins::IndexMixin
+      include Api::V1::Mixins::ShowMixin
+    end
+  end
+end
+
--- a/app/controllers/api/v3x0/service_credentials_controller.rb
+++ b/app/controllers/api/v3x0/service_credentials_controller.rb
@@ -1,0 +1,9 @@
+module Api
+  module V3x0
+    class ServiceCredentialsController < ApplicationController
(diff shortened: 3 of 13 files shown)


agrare:
@lpichler can you bump the openapi version so we can do a point release of the client?

lpichler:
> @lpichler can you bump the openapi version so we can do a point release of the client?

Done

lpichler:
I also added `ServiceCredentials` subcollections for `ServinceInstance, ServinceOffering, ServinceInstanceNode, ServinceInstanceNode`

and I did it like we have It for tagging, I used one controller(`Credentialings`) 
but `naming is hard` (I heard this quote from JoeR :) ). If you have better idea let me know. 

thanks!


cc @bdunne

app/controllers/api/v2x0/service_credential_types_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module V2x0
```
bdunne: Since we're following simver and v2.0 is already public, this should be in v2.1

gtanzillo:
I agree with Brandon's concerns...

I was not sure that the "tagging" pattern fits this use case and wanted to hear what others thought. I think it detracts more than it adds and it makes more sense to keep it simple.

Also, we'll need to have some tests here, as Brandon suggested.

slemrmartin:
@bdunne relations are similar to tags. We have 4 models, 4 join tables and 1 credentials (so 4x `has many, :through`) - tags have the same structure.

I would rather see paths like `service_offerings/:id/service_credentials` etc., but it's not normal has_many, so I don't know how it works with all the mixins

lpichler:
> * Does this need to be like tags with a special controller?  I'd prefer it to act more like a normal `has_many` relation if possible.

@bdunne  I guess you started with this ? any progress about it ? thanks!

app/controllers/api/v2x0/service_credential_types_controller.rb
bdunne: Do you mind basing this on v3.0 from https://github.com/RedHatInsights/topological_inventory-api/pull/291 ?

bdunne:
> > * Does this need to be like tags with a special controller?  I'd prefer it to act more like a normal `has_many` relation if possible.
> 
> @bdunne I guess you started with this ? any progress about it ? thanks!

Yes, this was merged in https://github.com/RedHatInsights/topological_inventory-api/pull/288

lpichler:
> > @bdunne I guess you started with this ? any progress about it ? thanks!
> 
> Yes, this was merged in #288

yes it works, 

PR updated 👍 

thanks

lpichler:
Client PR https://github.com/RedHatInsights/topological_inventory-api-client-ruby/pull/29

PR #285, opened 2020-02-27
Add v2.1

lpichler:
based on request in https://github.com/RedHatInsights/topological_inventory-api/pull/284#discussion_r384003101

cc @bdunne @abellotti

### DIFF
--- a/app/controllers/api/v2x1.rb
+++ b/app/controllers/api/v2x1.rb
@@ -1,0 +1,46 @@
+module Api
+  module V2x1
+    class RootController < ApplicationController
+      def openapi
+        render :json => ::Insights::API::Common::OpenApi::Docs.instance["2.1"].to_json
+      end
+    end
+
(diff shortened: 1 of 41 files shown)


app/controllers/api/v2x1.rb
```diff
@@ -0,0 +1,47 @@
+module Api
+  module V2x1
+    class RootController < ApplicationController
+      def openapi
+        render :json => ::Insights::API::Common::OpenApi::Docs.instance["2.1"].to_json
+      end
+    end
+
+    # Same behavior as api/v1.0/
```
bdunne: These should inherit from v2.0, not v1.0

lpichler: done 👍

bdunne: The comment is incorrect now.  It can be removed though.

lpichler: yes, removed, thanks 👍

PR #288, opened 2020-02-28
Allow subcollection requests to traverse has_many through relations

bdunne:
Load the primary instance and walk the declared has_many relation to get the results rather than starting at the class of the subcollection.


agrare:
@bdunne looks good, could you add a spec to cover this?

bdunne:
> @bdunne looks good, could you add a spec to cover this?

I was trying to come up with something reasonable, but it's very dependent on a request that we don't yet allow (there's no routes for these types of relations yet).

PR #289, opened 2020-03-05
Client generator: updating gem version

slemrmartin:
rake `client:generate` now updates `TopologicalInventoryApiClient::VERSION` regarding to latest API version (with patch version == always 0)

---

TPINVTRY-885

### DIFF
--- a/lib/tasks/client_generate.rake
+++ b/lib/tasks/client_generate.rake
@@ -39,7 +39,21 @@ class ClientGenerator
   end
 
   def generator_config
-    @generator_config ||= Pathname.new(Rails.root.join(".openapi_generator_config.json")).to_s
+    return @generator_config if @generator_config.present?
+
+    orig_config = Pathname.new(Rails.root.join(".openapi_generator_config.json")).to_s
+    json_settings = File.read(orig_config)
+
+    # Extend config by version used as client's Gem version
+    gem_version = "#{api_version}.0"
+    settings = JSON.parse(json_settings)
+    settings['gemVersion'] = gem_version
+    filepath = "/tmp/openapi_generator_#{gem_version}_#{Time.now.to_i}.json"
+    File.open(filepath, 'w') do |f|
+      f.write(settings.to_json)
+    end
+
+    @generator_config = Pathname.new(filepath).to_s
   end
 
   def openapi_file
--- a/lib/tasks/client_generate.rake
+++ b/lib/tasks/client_generate.rake
@@ -22,6 +22,10 @@ class ClientGenerator
     end.max
   end
 
+  def gem_version
+    @gem_version ||= ::Insights::API::Common::OpenApi::Docs.instance[api_version].version.to_s
+  end
+
   def generator_cli_jar
     @generator_cli_jar ||= begin
       jar_path = Pathname.new(Rails.root.join("public/doc/openapi-generator-cli-#{VERSION}.jar"))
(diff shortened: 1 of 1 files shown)


lib/tasks/client_generate.rake
```diff
@@ -39,7 +39,21 @@ class ClientGenerator
...
+    return @generator_config if @generator_config.present?
+
+    orig_config = Pathname.new(Rails.root.join(".openapi_generator_config.json")).to_s
+    json_settings = File.read(orig_config)
+
+    # Extend config by version used as client's Gem version
+    gem_version = "#{api_version}.0"
+    settings = JSON.parse(json_settings)
+    settings['gemVersion'] = gem_version
+    filepath = "/tmp/openapi_generator_#{gem_version}_#{Time.now.to_i}.json"
```
agrare: Can you use `Tempfile` instead of a constant path here?  It is possible for two `rake client:generate` tasks (even for two different gems) to overwrite each other's generator_config

lib/tasks/client_generate.rake
```diff
@@ -39,7 +39,21 @@ class ClientGenerator
...
 
   def generator_config
-    @generator_config ||= Pathname.new(Rails.root.join(".openapi_generator_config.json")).to_s
+    return @generator_config if @generator_config.present?
+
+    orig_config = Pathname.new(Rails.root.join(".openapi_generator_config.json")).to_s
+    json_settings = File.read(orig_config)
+
+    # Extend config by version used as client's Gem version
+    gem_version = "#{api_version}.0"
```
agrare: You should be able to get the full api_version from the openapi file, we were actually revving the last component of the path so `.0` isn't always correct

lib/tasks/client_generate.rake
slemrmartin: It's not possible, because Tempfile uses `unlink` method when removed by garbage collector and file is deleted.

Anyway `client:generate` is always run manually by developer, so (s)he probably don't run 2 rake task in the same `Time.now.to_i` time.

lib/tasks/client_generate.rake
slemrmartin: Do you mean from "info/version"? Which means I have to load and parse openapi.json file?

lib/tasks/client_generate.rake
agrare: > It's not possible, because Tempfile uses unlink method when removed by garbage collector and file is deleted.

Isn't that a good thing?  We don't want to leave cruft around in `/tmp` after the rake task finishes.

slemrmartin: ok, I'll make Tempfile instance variable, it should survive until the system("java...") call finishes

bdunne:
Since the same change was applied to sources-api, can this code be moved into the generator in `insights-api-common`? https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/open_api/generator.rb

slemrmartin:
> Since the same change was applied to sources-api, can this code be moved into the generator in `insights-api-common`? https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/open_api/generator.rb

@bdunne this class is used for `openapi:generate` rake task in both repos (and repo class inherits from it) and is intended to generate JSON, but this is ClientGenerator which shouldn't inherit from it (contains different logic and methods). I think some refactoring in insights-api-common is needed first.

PR #291, opened 2020-03-11
Bumping to v3.0 - PaginatedResponseV2

abellotti:
Bumping up Topological Inventory Api to v3.0
- Support for PaginatedResponseV2 for both REST and GraphQL
- Adding v3 specs

Fixes: TPINVTRY-864

### DIFF
--- a/app/controllers/api/v3x0.rb
+++ b/app/controllers/api/v3x0.rb
@@ -1,0 +1,9 @@
+module Api
+  module V3x0
+    class RootController < ApplicationController
+      def openapi
+        render :json => ::Insights::API::Common::OpenApi::Docs.instance["3.0"].to_json
+      end
+    end
+  end
+end
--- a/app/controllers/api/v3x0/mixins/index_mixin.rb
+++ b/app/controllers/api/v3x0/mixins/index_mixin.rb
@@ -1,0 +1,18 @@
+module Api
+  module V3x0
+    module Mixins
+      module IndexMixin
+        def index
+          raise_unless_primary_instance_exists
+          render :json => Insights::API::Common::PaginatedResponseV2.new(
+            :base_query => scoped(filtered.where(params_for_list)),
+            :request    => request,
+            :limit      => pagination_limit,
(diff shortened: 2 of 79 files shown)


lpichler:
@abellotti 

- when this will be merged, I guess we don't need 2.1 https://github.com/RedHatInsights/topological_inventory-api/pull/285 ?

app/controllers/api/v3x0.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module V3x0
+    class RootController < ApplicationController
+      def openapi
```
lpichler: @abellotti 

Is here reason to not leverage controllers from previous version like it is here 
https://github.com/lpichler/topological_inventory-api/blob/b1078f434b15e7db041721fd17805f85150d22e0/app/controllers/api/v2x0.rb#L10
?
and to have in v3.0 controllers  just `Mixins::IndexMixin` ?

thanks!

bdunne: @lpichler We need to pull in the `Api::V3x0::Mixins::IndexMixin`, not the one from `V1x0`

app/controllers/api/v3x0/mixins/index_mixin.rb
```diff
@@ -0,0 +1,18 @@
+module Api
+  module V3x0
+    module Mixins
+      module IndexMixin
+        def index
+          raise_unless_primary_instance_exists
+          render :json => Insights::API::Common::PaginatedResponseV2.new(
+            :base_query => scoped(filtered.where(params_for_list)),
```
bdunne: @abellotti Does this need to inherit from the `V1x0` mixin for the `scoped` method?

bdunne: Nevermind, I see, the controller inheritance is taking care of it.

app/controllers/api/v3x0.rb
lpichler: 👍 👍

PR #293, opened 2020-03-25
Enabling filtering by association attribute for the REST API.

abellotti:
- [x] Depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/159

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -120,7 +120,11 @@ class ApplicationController < ActionController::API
   end
 
   def filtered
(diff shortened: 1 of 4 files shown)


bdunne:
@abellotti Can you also require the latest insights-api-common v3.9?

abellotti:
Requires build 3.10 of insighs-api-common to pull in: https://github.com/RedHatInsights/insights-api-common-rails/pull/176

abellotti:
Ping @syncrou, just rebased and repushed.  Please take a 👀  Thanks.

syncrou:
@mkanoor could you take a look at this?

abellotti:
Just needed to do what was done with Sources https://github.com/RedHatInsights/sources-api/pull/204/files  so that we're consistent in what we support in the REST API.

PR #301, opened 2020-04-30
Rename task_id key to id

syncrou:
Renames the task_id key to id for all task update events.

This brings the operations events inline with persister events that were already including an 'id' key.

This PR is required for https://github.com/RedHatInsights/catalog-api-minion/pull/34 to work correctly.


syncrou:
@gmcculloug @lindgrenj6 Please review

cc - @agrare

PR #302, opened 2020-05-05
Store whole forwardable_headers in Task

slemrmartin:
**Issue** https://github.com/RedHatInsights/topological_inventory-persister/issues/58

Insights::API::Common::Request.current_forwardable is saved in Task instead of x-rh-insights-request-id. To be able to resend it through Kafka in persister

---

* [x] **depends on** https://github.com/RedHatInsights/topological_inventory-core/pull/197
  - **requires** RubyGems v1.1.4

### DIFF
--- a/public/doc/openapi-3-v3.0.json
+++ b/public/doc/openapi-3-v3.0.json
@@ -11307,10 +11307,6 @@
             "readOnly": true,
             "type": "string"
           },
-          "forwardable_headers": {
-            "example": "Insights::API::Common::Request.current_forwardable for later usage",
-            "type": "object"
-          },
           "id": {
             "$ref": "#/components/schemas/ID"
           },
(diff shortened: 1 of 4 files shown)


syncrou:
@lindgrenj6 please review

public/doc/openapi-3-v3.0.json
```diff
@@ -11307,6 +11307,10 @@
             "readOnly": true,
             "type": "string"
           },
+          "forwardable_headers": {
```
gmcculloug, line 11310: I do not think `forwardable_headers` should be exposed through the API.  We only need it for internal processing.  cc @syncrou @mkanoor

syncrou: I agree.  Seems we shouldn't expose forwardable headers in the openapi spec

mkanoor: Yes its internal and should not be exposed.

slemrmartin: I can remove it from this PR, but generally the `openapi:generate` rake task must be extended to support skipping

slemrmartin: TPINVTRY-932

PR #324, opened 2020-10-27
Remove `config.log_level = :debug` from environments/production.rb

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10286

This is why the log level is ALWAYS DEBUG in production mode. 

By removing this line it will respect what we set in `config/application.rb` and thus the ENV var in the template.


slemrmartin (approved):
:+1:

syncrou (approved):
LGTM 👍

PR #325, opened 2020-10-29
Move `messaging_client` to a singleton class to prevent threading issues

lindgrenj6:
Looking at production after pushing the `rdkafka` changes we are running out of threads just like in Ingress API, turns out that the messaging_client was an instance method that created a new client every time and never closed it, thus leading to running out of threads eventually.

This fix just moves it to a singleton class.

### DIFF
--- a/lib/topological_inventory/api/messaging.rb
+++ b/lib/topological_inventory/api/messaging.rb
@@ -1,0 +1,16 @@
+module TopologicalInventory
+  module Api
+    module Messaging
+      def self.client
+        require "manageiq-messaging"
+
+        @client ||= ManageIQ::Messaging::Client.open({
+          :protocol => :Kafka,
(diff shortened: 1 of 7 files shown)


lib/topological_inventory/api/messaging.rb
```diff
@@ -0,0 +1,18 @@
+module TopologicalInventory
+  module Api
+    class Messaging
```
slemrmartin, line 3: it's a class there, but Module in Sources API, pls unify it

slemrmartin (changes requested):

lib/topological_inventory/api/messaging.rb
lindgrenj6: Doen

slemrmartin (approved):
:+1: LGTM

